PR #6, opened 2018-12-11
Fix panic because yaml assets path is wrong

wongma7:
Made a series of dumb mistakes/assumptions, I moved yamls without changing the code, this should fix the operator insta-panicking


bertinatto:
/lgtm

(assuming happy tests)

PR #9, opened 2019-01-11
Update Makefile to use FIRST_GOPATH

sosiouxme:
GOPATH does not have to be a single path. Downstream, it isn't.


sosiouxme:
@wongma7

PR #14, opened 2019-02-20
Handle unsupported platform errors gracefully

gnufied:
Unsupported platform isn't really a fatal error. We are just not going to install a storageclass. This fixes deployment on libvirt.

### DIFF
--- a/pkg/controller/clusterstorage/clusterstorage_controller.go
+++ b/pkg/controller/clusterstorage/clusterstorage_controller.go
@@ -192,34 +192,41 @@ func (r *ReconcileClusterStorage) syncStatus(clusterOperator *configv1.ClusterOp
 		Status: configv1.ConditionFalse,
 	}
 	v1helpers.SetStatusCondition(&clusterOperator.Status.Conditions, notProgressing)
+	var message string
 
 	// if error is anything other than unsupported platform, we are failing
-	if err != nil && err != unsupportedPlatformError {
-		failing := configv1.ClusterOperatorStatusCondition{
-			Type:    configv1.OperatorFailing,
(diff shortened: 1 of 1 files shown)


smarterclayton:
Yeah, as a general rule all operators everywhere must gracefully tolerate unrecognized platforms.  See https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go#L34 for a quick description and 
https://github.com/openshift/installer/pull/1112 for more details

jsafrane:
This code lgtm, however, should the operator tell user somehow that it's running correctly but did not do anything useful on an unknown platform? *"I see the operator running, it does not report any error, so where is my default storage class?"*

pkg/controller/clusterstorage/clusterstorage_controller.go
```diff
@@ -192,7 +193,8 @@ func (r *ReconcileClusterStorage) syncStatus(clusterOperator *configv1.ClusterOp
 	}
 	v1helpers.SetStatusCondition(&clusterOperator.Status.Conditions, notProgressing)
 
-	if err != nil {
+	// if error is anything other than unsupported platform, we are failing
+	if err != nil && err != unsupportedPlatformError {
```
bertinatto: +1: it could be useful to have a log entry if `err == unsupportedPlatformError`.

gnufied: fixeds

gnufied:
@jsafrane added reason field if unsupported platform is detected. PTAL

wongma7:
reason should be camelcase. maybe put it in Message instead?

gnufied:
@wongma7 fixed.

PR #16, opened 2019-03-01
version: Set the cluster operator version based on the payload

smarterclayton:
The new contract with operators is that we report the version when
we reach "level" (have rolled out the version we got handed in the
payload) and are available, which ensures that the CVO can wait
until it sees us completely roll out the payload. This makes upgrade
status much clearer.

See https://github.com/openshift/cluster-image-registry-operator/pull/219 for more context, we are rolling this out across all operators.


smarterclayton:
@gnufied

Issue #19, opened 2019-03-19
[libvirt] CVO is not able to update storage operator and timeout.


praveenkumar:
I tried latest master of installer and looks like CVO time out during the `update of storage operator` which never finish and finally error out.

```console
$ openshift-install version
openshift-install unreleased-master-577-g9777df02d2a0dc287bb520c5ea2f409499c28eca
built from commit 9777df02d2a0dc287bb520c5ea2f409499c28eca

$ oc get clusterversion
NAME      VERSION                           AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.0.0-0.alpha-2019-03-19-045826   False       True          27m     Working towards 4.0.0-0.alpha-2019-03
-19-045826: 99% complete

$ oc logs cluster-version-operator-8dc5dc7fb-xslw4 -n openshift-cluster-version | grep error
[...]
E0319 09:35:19.656479       1 task.go:58] error running apply for clusteroperator "monitoring" (200 of 309): Cluster operator monitoring is still updating
E0319 09:41:18.089625       1 task.go:58] error running apply for clusteroperator "storage" (234 of 309): Cluster operator storage is still updating
E0319 09:47:20.090892       1 task.go:58] error running apply for clusteroperator "storage" (234 of 309): Cluster operator storage is still updating
E0319 09:53:24.092701       1 task.go:58] error running apply for clusteroperator "storage" (234 of 309): Cluster operator storage is still updating
E0319 10:00:22.223533       1 task.go:58] error running apply for clusteroperator "storage" (234 of 309): Cluster operator storage is still updating
E0319 10:06:23.224944       1 task.go:58] error running apply for clusteroperator "storage" (234 of 309): Cluster operator storage is still updating

$ oc logs cluster-storage-operator-6d48465ddc-nf7p2 -n openshift-cluster-storage-operator
{"level":"info","ts":1552987778.9392104,"logger":"cmd","msg":"Version of operator-sdk: v0.4.0"}
{"level":"info","ts":1552987778.9575768,"logger":"leader","msg":"Trying to become the leader."}
{"level":"info","ts":1552987780.0816255,"logger":"leader","msg":"No pre-existing lock was found."}
{"level":"info","ts":1552987780.094624,"logger":"leader","msg":"Became the leader."}
{"level":"info","ts":1552987780.494817,"logger":"cmd","msg":"Registering Components."}
[...]

PR #20, opened 2019-03-19
Do not re-queue if error is unsupported

gnufied:
I think even though ClusterOperator is being created correctly, we need to stop the operation from requeing.  It should fix - https://github.com/openshift/cluster-storage-operator/issues/19

cc @wongma7 @bertinatto


bertinatto:
/lgtm
(but @wongma7 may want to take a look as well)

gnufied:
putting on hold for mattew to take a look.

/hold

wongma7:
why would it fix 19? Even if it requeues it should set the status to Available: true

wongma7:
/retest

we should merge this but I don't know about #19

gnufied:
Yeah I think #19 is not caused by error from unsupported platform.  I will remove the hold.

/hold cancel

PR #21, opened 2019-03-19
Set versions if unsupported platform but Available

wongma7:


pkg/controller/clusterstorage/clusterstorage_controller.go
```diff
@@ -188,7 +188,7 @@ func (r *ReconcileClusterStorage) syncStatus(clusterOperator *configv1.ClusterOp
 	// we set versions if we are "available" to indicate we have rolled out the latest
 	// version of the cluster storage object
 	if releaseVersion := os.Getenv("RELEASE_VERSION"); len(releaseVersion) > 0 {
-		if err == nil {
+		if err == nil || err == unsupportedPlatformError {
 			clusterOperator.Status.Versions = []configv1.OperandVersion{{Name: "operator", Version: releaseVersion}}
```
gnufied: yeah I arrived on same conclusion, we aren't setting versions at all on unsupported platforms. So compared to kube-controller-manager operator:

```
  "versions": [
            {
                "name": "operator",
                "version": "4.0.0-0.alpha-2019-03-19-111025"
            },
            {
                "name": "kube-controller-manager",
                "version": "1.12.4"
            }

```

wongma7:
yes this is it: "There MUST be a version with the name operator, which is watched by the CVO to know if a cluster operator has achieved the new level" https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#version.
/retest

PR #22, opened 2019-03-20
Set progressing status just before trying to create SC

wongma7:
fix bz 1688582


wongma7:
@openshift/storage pls review.
operator is supposed to set Progressing before available/failing. otherwise user does not see 'lastTransitionTime' as there is no transition available->available, even though the version may actually have changed

pkg/controller/clusterstorage/clusterstorage_controller.go
```diff
@@ -186,7 +192,58 @@ func (r *ReconcileClusterStorage) Reconcile(request reconcile.Request) (reconcil
...
+	notProgressing = configv1.ClusterOperatorStatusCondition{
+		Type:   configv1.OperatorProgressing,
+		Status: configv1.ConditionFalse,
+	}
+)
+
+// setStatusProgressing sets Available=false;Failing=false;Progressing=true
+// we set "progressing" if the cluster operator's version is not the latest
+// and we are about to try to roll it out
+func (r *ReconcileClusterStorage) setStatusProgressing(clusterOperator *configv1.ClusterOperator) error {
```
gnufied: Can we write some unit tests for this stuff?

wongma7: done. will write one for syncStatus in follow up

gnufied:
Failures here look unrelated. (storage flakes... :( )

/lgtm

PR #25, opened 2019-03-29
Add vsphere default storageclass

wongma7:
We will need to manually test this with a cluster where the cloudprovider is configured.

### DIFF
--- a/assets/vsphere.yaml
+++ b/assets/vsphere.yaml
@@ -1,0 +1,9 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: thin
+  annotations:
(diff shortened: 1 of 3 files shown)


wongma7:
re: previous discussion about datastore, we should be okay to leave it blank. Ideally the user will configure it in their cloudprovider and won't even need to create a new storageclass if there is a specific datastore they want to use:
"datastore: The user can also specify the datastore in the StorageClass. The volume will be created on the datastore specified in the storage class, which in this case is VSANDatastore. This field is optional. If the datastore is not specified, then the volume will be created on the datastore specified in the vSphere config file used to initialize the vSphere Cloud Provider."

PR #28, opened 2019-04-19
Add azure default storageclass

juan-lee:


assets/azure.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: managed-premium-retain
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"
+provisioner: kubernetes.io/azure-disk
+reclaimPolicy: Retain
```
wongma7: I think it should be Delete, i.e. unset.

assets/azure.yaml
```diff
@@ -0,0 +1,11 @@
...
+parameters:
+  kind: Managed
+  storageaccounttype: Premium_LRS
```
wongma7: what if the VM is standard or unmanaged? Ideally the default "standard" class won't have any parameters set, so that it works with every cluster, like https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/storage-class/azure/default.yaml. Can we safely assume that openshift will only be installed on clusters where VMs are premium+managed?

assets/azure.yaml
wongma7: this is only to be consistent with other clouds. Whether admin wants their disks automatically cleaned up is a matter of preference/policy I think, so staying consistent would be better IMO.

assets/azure.yaml
juan-lee: @wongma7 Azure's managed disk offering works with all of our VMs and is our recommended default. Standard storageaccounttype is meant for testing/entry level prod workloads where Premium is what we recommend for our customers for production use. I also checked and managed-premium is the closest equivalent to the aws default storage class.

assets/azure.yaml
juan-lee: removed

wongma7:
/lgtm

thanks

PR #32, opened 2019-05-05
Implement STOR-213 - Default Encryption for GP2

cuppett:
In 3.x, PV encryption is enabled by default in the gp2 storage class. On a recent OCP 4.1 install, I see that the default storage class does not have it enabled:

kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: gp2
  ...
  annotations:
    storageclass.kubernetes.io/is-default-class: 'true'
  ownerReferences:
    ...
provisioner: kubernetes.io/aws-ebs
parameters:
  type: gp2
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer

We would like to see this default storage class be encrypted by default to mirror the behavior of OSD 3.x clusters.


wongma7:
what is osd, dedicated? openshift-ansible defaulted to encrypted: false. https://github.com/openshift/openshift-ansible/blob/release-3.11/roles/openshift_default_storage_class/defaults/main.yml#L12

cuppett:
Yes. Dedicated. They were changing after install for new customers. I would
expect most others would as well.

In addition, we use encrypted AMIs now in AWS via the OCP 4.1 installer.
This change ensures additional EBS volumes default to encrypted as well.

wongma7:
OK, sgtm, let's get this change in before code freeze then... Please run `make generate` in the repo root else the change in yaml won't apply.
/approve

cuppett:
Pushed the bin change as well. Thanks!

PR #34, opened 2019-05-12
Enable AWS EBS volume expansion by default

cblecker:
Add the allowVolumeExamsion option to to the default gp2 storage class, allowing for volumes to be expanded automatically after they are created.

https://docs.openshift.com/container-platform/4.1/storage/expanding-persistent-volumes.html


bertinatto:
I suppose we don't add the volume expansion to the standard storage class because the feature is beta in 4.1. What do you think, @gnufied?

gnufied:
so far the reasoning for not enabling resizing in default storageclass wasn't because feature was in beta but because for AWS in particular, the volume resizing quota is pretty low. Like you can only resize same volume every 6 minutes or something. I dunno what the global quota looks like, but I would assume it would be pretty low too.

Enabling it in deault SC will allow all users tor resize their PVCs. @cblecker is that what we want?

cblecker:
@gnufied I'm not sure. I wasn't aware of that restriction. I know from an SRE point of view, the flexibility of expanding volumes out of the box is appealing, but I'm not sure of drawbacks. There's also an aos-devel thread about this.

cblecker:
To update, SRE is discussing this internally.

cblecker:
@gnufied SRE supports this change, so I think we should be safe to merge. If possible, we'd like this for 4.1 GA.

PR #35, opened 2019-05-16
manifests: add resource requests to operator pod

sjenning:
in service to https://github.com/openshift/origin/pull/22787

### DIFF
--- a/manifests/02-deployment.yaml
+++ b/manifests/02-deployment.yaml
@@ -52,3 +52,7 @@ spec:
                   fieldPath: metadata.name
             - name: OPERATOR_NAME
               value: "storage"
+          resources:
+            requests:
+              cpu: 10m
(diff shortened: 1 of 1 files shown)


gnufied:
How did we determine CPU numbers btw? I am just curious.

sjenning:
@gnufied just looked at the Prometheus metrics for the pod

```
sort_desc(sum (rate(container_cpu_usage_seconds_total{namespace="openshift-cluster-storage-operator"}[1m])) by (pod_name))
sort_desc(sum (container_memory_usage_bytes{namespace="openshift-cluster-storage-operator"}) by (pod_name))
```
cpu
memory

sjenning:
memory probably could have been `30Mi` but the main thing is to set something.  Something is better than nothing.

PR #41, opened 2019-07-16
Bug 1727044: Set missing Upgradeable condition

bertinatto:
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1727044


bertinatto:
CC @openshift/storage

PR #42, opened 2019-07-24
Bug: 1731059: Use topology aware dynamic provisioning in the clouds

jsafrane:
AWS already uses `volumeBindingMode: WaitForFirstConsumer`, the others should do the same to have consistent behavior and benefits of better scheduling.

I am not sure about vsphere, does it have topology? Even if it had not, it makes sense to me to have the default storage classes consistent.

cc @openshift/storage 

/assign @gnufied 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1731059


assets/vsphere.yaml
```diff
@@ -5,5 +5,6 @@ metadata:
   annotations:
     storageclass.kubernetes.io/is-default-class: "true"
 provisioner: kubernetes.io/vsphere-volume
+volumeBindingMode: WaitForFirstConsumer
```
gnufied: vsphere does not support topology aware dynamic provisioning. Should we verify nothing breaks if this option is enabled?

jsafrane:
Removed vSphere changes, it does not support topology.

PR #43, opened 2019-07-30
Adding the default GCP StorageClass.

huffmanca:
This adds the default StorageClass for GCP environments.

This is for OS 4.2.


assets/gcp.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: gce-standard
```
gnufied: We should call this just `standard` to be inline with old openshift-ansible.

huffmanca: This has been updated this to be `standard`.

PR #45, opened 2019-08-15
Bug 1741465: Enable volume expansion on platforms where is supported

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1741465


jsafrane:
/lgtm
/retitle Bug 1741465: Enable volume expansion on platforms where is supported

PR #47, opened 2019-09-12
Bug 1751641: Enables updating of StorageClass if a change is detected.

huffmanca:
Updates the ClusterStorageOperator to check for changes made to the default StorageClass assets. If changes are found in either AllowVolumeExpansion or MountOptions, update the default StorageClass to use the new values.

This change also prevents users from adjusting these attributes from the ones used in the assets. We should likely document that restriction, and advise users to create a new storage class if changes are needed.

### DIFF
--- a/pkg/controller/clusterstorage/clusterstorage_controller.go
+++ b/pkg/controller/clusterstorage/clusterstorage_controller.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"reflect"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-storage-operator/pkg/generated"
@@ -145,7 +146,7 @@ func (r *ReconcileClusterStorage) Reconcile(request reconcile.Request) (reconcil
 	}
 
 	// Define a new StorageClass object
-	sc, err := newStorageClassForCluster(instance)
+	newSCFromFile, err := newStorageClassForCluster(instance)
 	if err != nil {
 		_ = r.syncStatus(clusterOperatorInstance, err)
 		// requeue only if platform is supported
@@ -154,10 +155,10 @@ func (r *ReconcileClusterStorage) Reconcile(request reconcile.Request) (reconcil
 		}
 		return reconcile.Result{}, nil
 	}
-	clusterOperatorInstance.Status.RelatedObjects = getRelatedObjects(sc)
+	clusterOperatorInstance.Status.RelatedObjects = getRelatedObjects(newSCFromFile)
 
 	// Set the clusteroperator to be the owner of the SC
-	ocontroller.EnsureOwnerRef(sc, metav1.OwnerReference{
+	ocontroller.EnsureOwnerRef(newSCFromFile, metav1.OwnerReference{
 		APIVersion: "v1",
 		Kind:       "clusteroperator",
 		Name:       clusterOperatorName,
@@ -165,11 +166,11 @@ func (r *ReconcileClusterStorage) Reconcile(request reconcile.Request) (reconcil
 	})
 
 	// Check if this StorageClass already exists
-	found := &storagev1.StorageClass{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: sc.Name, Namespace: corev1.NamespaceAll}, found)
+	existingSC := &storagev1.StorageClass{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: newSCFromFile.Name, Namespace: corev1.NamespaceAll}, existingSC)
 	if err != nil && apierrors.IsNotFound(err) {
-		reqLogger.Info("Creating a new StorageClass", "StorageClass.Name", sc.Name)
-		err = r.client.Create(context.TODO(), sc)
+		reqLogger.Info("Creating a new StorageClass", "StorageClass.Name", newSCFromFile.Name)
+		err = r.client.Create(context.TODO(), newSCFromFile)
 		if err != nil {
 			_ = r.syncStatus(clusterOperatorInstance, err)
 			return reconcile.Result{}, err
@@ -183,8 +184,34 @@ func (r *ReconcileClusterStorage) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
-	// StorageClass already exists - don't requeue
-	reqLogger.Info("Skip reconcile: StorageClass already exists", "StorageClass.Name", found.Name)
+	// Check to see if modifications have been made to the StorageClass attributes
+	comparisonSC := newSCFromFile.DeepCopy()
+	// Copy over the ObjectMeta, which includes the annotations and labels
+	comparisonSC.ObjectMeta = existingSC.ObjectMeta
+
+	// Define a default ReclaimPolicy for comparison
+	if comparisonSC.ReclaimPolicy == nil {
+		deletePolicy := corev1.PersistentVolumeReclaimDelete
+		comparisonSC.ReclaimPolicy = &deletePolicy
+	}
+
+	// If a change has been detected, update the StorageClass.
+	if !reflect.DeepEqual(comparisonSC, existingSC) {
+		reqLogger.Info("StorageClass already exists and needs to be updated", "StorageClass.Name", existingSC.Name)
+
+		err = r.client.Update(context.TODO(), comparisonSC)
+		if err != nil {
+			_ = r.syncStatus(clusterOperatorInstance, err)
+			return reconcile.Result{}, err
+		}
+
+		// StorageClass updated successfully - don't requeue
+		_ = r.syncStatus(clusterOperatorInstance, nil)
+		return reconcile.Result{}, nil
+	}
+
+	// StorageClass already exists and doesn't need to be updated - don't requeue
+	reqLogger.Info("Skip reconcile: StorageClass already exists", "StorageClass.Name", existingSC.Name)
 	_ = r.syncStatus(clusterOperatorInstance, nil)
 	return reconcile.Result{}, nil
 }
--- a/pkg/controller/clusterstorage/clusterstorage_controller_test.go
+++ b/pkg/controller/clusterstorage/clusterstorage_controller_test.go
@@ -3,15 +3,19 @@ package clusterstorage
 import (
 	"context"
 	"os"
+	"reflect"
+	"strconv"
 	"testing"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-storage-operator/pkg/apis"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 func TestSetStatusProgressing(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/controller/clusterstorage/clusterstorage_controller.go
```diff
@@ -183,7 +184,34 @@ func (r *ReconcileClusterStorage) Reconcile(request reconcile.Request) (reconcil
...
+	needUpdate := false
+	if sc.AllowVolumeExpansion != found.AllowVolumeExpansion {
+		found.AllowVolumeExpansion = sc.AllowVolumeExpansion
+		needUpdate = true
+	}
+
+	if !reflect.DeepEqual(sc.MountOptions, found.MountOptions) {
+		found.MountOptions = sc.MountOptions
+		needUpdate = true
+	}
```
bertinatto: If I understand correctly, if the user has `mountOptions` set, we overwrite it with an empty value (since `mountOptions` is not set in our yaml file for 4.2). The same happens for `allowVolumeExpansion`.

However, if the user has deliberately chosen to set `volumeBindingMode` to `Immediate` in 4.1, we won't make it `WaitForFirstConsumer`, which is also the default for 4.2.  In other words, we enforce updating **some** fields, but not **others**.

I think ideally we should make the operator the "owner" of this SC, always updating it to the version we ship in the yaml file.

bertinatto: We could achieve that by:

1) deleting and re-creating the storageClass when they differ. I'm not sure if this is acceptable (a user might create a PVC during the re-creation process).

2) Always update the storageClass, but making sure we don't overwrite fields that cannot be changed (e.g., `parameters`). This is only to cover cases where the user deleted the SC in 4.1 and re-created it changing only a few fields (like `parameters`).

@jsafrane @gnufied: thoughts?

jsafrane: I agree with @bertinatto, the operator should probably own the storage class it crates and try to overwrite user changes in the fields that can be changed.

We should document that the default storage class is owned by OpenShift and users should create their own one if they don't like ours.

huffmanca: To change `volumeBindingMode` we'd have to recreate the class, correct? Because the field is immutable from https://github.com/kubernetes/kubernetes/blob/8126201b73b3fa52bd376fc98cb00637af493b72/pkg/apis/storage/validation/validation.go#L75 ?

I've moved this to an internal function and updated the PR to update the 3 fields that I currently see we can change.

I definitely agree that we'll need a documentation PR as well.

gnufied: if we delete and re-create the SC then we will lose annotation (set by user) that disables this SC to be default SC and it should be possible to disable this SC as default SC. 

If we are going to delete and re-create the SC then at very minimum we should retain previous value of `storageclass.kubernetes.io/is-default-class` annotation.  But can a case be made that why should `storageclass.kubernetes.io/is-default-class` be treated as special?

I guess this is why all objects managed by an operator should have `UnsupportedConfigOverrides` so as users don't directly modify objects created by the operator but by not having a CR or api for this, makes it harder. 


For now - my vote is delete and re-create the SC but keep default SC annotation to previous value.

huffmanca: I've updated the PR to check for differences between the configurable fields, and if one is found to delete and recreate the StorageClass. This should now keep the annotations from the previous StorageClass.

pkg/controller/clusterstorage/clusterstorage_controller.go
```diff
@@ -336,6 +363,21 @@ func newStorageClassForCluster(infrastructure *configv1.Infrastructure) (*storag
...
+		*existingClass.AllowVolumeExpansion != *sc.AllowVolumeExpansion ||
+		!reflect.DeepEqual(existingClass.MountOptions, sc.MountOptions) ||
+		!reflect.DeepEqual(existingClass.AllowedTopologies, sc.AllowedTopologies) {
+		needUpdate = true
+	}
+
+	return needUpdate
+}
+
 func getRelatedObjects(sc *storagev1.StorageClass) []configv1.ObjectReference {
```
gnufied: we typically do this by replacing a field that can be changed (annotations in this case) and then comparing the two objects. so something like:

```
newSCCopy := newSC.DeepCopy()
newSCCopy.Annotation = oldSC.Annotation

if !reflect.DeepEqual(newSCCopy, oldSC) {
  returns true
}
return false
````

Comment (gnufied) [reply pkg/controller/clusterstorage/clusterstorage_controller.go]: ah this comparsion might not work for SC because it does not have spec. I do not like having to hardcode fields and we will need some tests.

Comment (gnufied) [reply pkg/controller/clusterstorage/clusterstorage_controller.go]: I think we should preserve all fields from `ObjectMeta` type rather than just annotations. So something like:

```golang
newsc.ObjectMeta =oldSC.ObjectMeta
```

should work. however before we merge this, @bertinatto what was your concern about SC being deleted while PVC was being provisioned? Shouldn't the PVC be pending and start getting provisioned as soon as new SC comes back?

huffmanca: I've used @gnufied's method to copy the old SC and examine the ObjectMeta. In addition, I have to check to see if the ReclaimPolicy is nil, as it's not defined in the `assets/*.yaml` file, and map this to the default value.

huffmanca:
I've currently in progress on adding a test for the Reconcile method. As of right now the SC in the client doesn't appear to be updated despite calling Reconcile. I'll need to investigate this further to determine if I'm calling this incorrectly in the test.

Manual testing of the changes has been successful, with the SC being updated to the contents of the `assets/*.yaml` file while keeping annotations and labels.

pkg/controller/clusterstorage/clusterstorage_controller.go
bertinatto: > Shouldn't the PVC be pending and start getting provisioned as soon as new SC comes back?

I just tested this and yes, it behaves like you said. I think we're good re-creating the SC.

pkg/controller/clusterstorage/clusterstorage_controller.go
```diff
@@ -183,7 +184,47 @@ func (r *ReconcileClusterStorage) Reconcile(request reconcile.Request) (reconcil
...
+		needUpdate = true
+	}
+
+	// If a change has been detected, delete and recreate the StorageClass, preserving annotations
+	// so that the default StorageClass value is not overridden.
+	if needUpdate {
+		reqLogger.Info("StorageClass already exists and needs to be recreated", "StorageClass.Name", found.Name)
+		annotations := found.Annotations
+		labels := found.Labels
+		err = r.client.Delete(context.TODO(), found)
```
jsafrane: If the operator wants to change just the fields that are changeable, the operator should do `Update`.

Re-create is dangerous, because it leaves the cluster without any default storage class for a short while and newly created PVCs will not get it. It will leave the PVCs without any storage class until user notices it and does something about it manually. I would even say that we won't re-create the storage class at all and just report an error.

gnufied: Discussed this in our standup call - it looks like choosing between two bad options. If we don't re-create SC then that would mean we could *never* update things like SC's parameter or binding mode.

On the other hand if we delete and re-create SC then there could be a short period during upgrade process where no SC will be found.

jsafrane: After-meeting thoughts:

I talked to @vrutkovs, currently it seems that major update (e.g. 4.2->4.3) will not be automatic. We can expect that admin reads release notes and can see that they should not create PVCs during the update. We're going to modify immutable SC fields only when really really necessary (if ever) and only during such major update. In that case, I would say that documentation is enough ("check your vSphere PVCs after update").

During minor update (or during normal operation):

- If user changes fields of the storage class, the operator should `Update()` it.
- What if user re-creates the storage class (to change immutable field)? We should clearly document that the storage class created by the operator is ours and users should not change it except for the default annotation. But then? Should the operator re-create (with an event? where? it would be great to get a metrics / alert for this). Now that's clear user error I am open to re-create the SC.

vrutkovs: I don't think exact Cincinnati schema for major updates has been prepared yet (cc @steveeJ @wking), so its unlikely to happen in a full automatic way in any case

steveeJ: I'm not aware of any special casing for major updates, so I assume that if the cluster is configured to automatically apply updates a major one will be no different from a minor one. Of course we could attach metadata to any release and teach the the CVO to behave differently when seeing such an update.

huffmanca: For now I've updated the Operator so that it will perform an Update if it notices differences between the two StorageClass entities. If it errors attempting to update an invalid field, then the Operator will enter a degraded state and log the error.

Marking this discussion as resolved for now.

pkg/controller/clusterstorage/clusterstorage_controller.go
```diff
@@ -183,7 +184,36 @@ func (r *ReconcileClusterStorage) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
-	// StorageClass already exists - don't requeue
+	// Check to see if modifications have been made to the StorageClass attributes
+	needUpdate := false
+	newSCC := sc.DeepCopy()
```
gnufied: SCC is a kinda not a great variable name because it has other meaning in context of openshift.

huffmanca: I updated this to be `comparisonSC`.

pkg/controller/clusterstorage/clusterstorage_controller.go
```diff
@@ -183,7 +184,36 @@ func (r *ReconcileClusterStorage) Reconcile(request reconcile.Request) (reconcil
...
+	}
+
+	if !reflect.DeepEqual(comparisonSC, found) {
+		needUpdate = true
+	}
+
+	// If a change has been detected, update the StorageClass.
+	if needUpdate {
+		reqLogger.Info("StorageClass already exists and needs to be updated", "StorageClass.Name", found.Name)
+		err = r.client.Update(context.TODO(), sc)
```
jsafrane: Does `sc` has the default annotation? When user changes a parameter *and* makes the SC as non-default, this `Update` makes it back default.

You should copy the annotation (or no annotation) from `found`.

bertinatto: I think you want to use `comparisonSC` rather than `sc`, right?

pkg/controller/clusterstorage/clusterstorage_controller.go
```diff
@@ -183,7 +184,36 @@ func (r *ReconcileClusterStorage) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
-	// StorageClass already exists - don't requeue
+	// Check to see if modifications have been made to the StorageClass attributes
+	needUpdate := false
+	comparisonSC := sc.DeepCopy()
+	comparisonSC.ObjectMeta = found.ObjectMeta
```
bertinatto: nit: perhaps `[old|existing]SC` and `newSC` would be a better representation for the storageClasses?

pkg/controller/clusterstorage/clusterstorage_controller_test.go
```diff
@@ -93,3 +98,85 @@ func TestSetStatusProgressing(t *testing.T) {
 		})
 	}
 }
+
+func TestReconcile(t *testing.T) {
+	volumeExpansion := false
+	newStorageClass := resourceread.ReadStorageClassV1OrDie(generated.MustAsset("assets/aws.yaml"))
```
bertinatto: nit: IMO the test could be more representative if you make `newStorageClass` a field in the anonymous struct below.

That way we could add more test cases with different variations, plus everything would be more deterministic (e.g., currently the test changes if we change the `assets/aws.yaml` file).

pkg/controller/clusterstorage/clusterstorage_controller.go
huffmanca: Changed this call `Update` on `comparisonSC`, which copies over the ObjectMeta from earlier. Since we're currently using `Update`, instead of `Create`, this seems to apply without issue and keeps the annotations and labels.

huffmanca: Previously, when we were using `Create`, this failed because we had previously copied over the `ObjectMeta`. Updating this line to `comparisonSC` updates the full ObjectMeta without issue. I went ahead and made this change.

pkg/controller/clusterstorage/clusterstorage_controller.go
huffmanca: I went through the method and updated the variables names as follows:

* found - has now been set to `existingSC`
* sc - has now been set to `newSCFromFile`

pkg/controller/clusterstorage/clusterstorage_controller_test.go
huffmanca: Updated the test to fit in with the typical method of testing, and included a helper method to obtain the fake storage class.

pkg/controller/clusterstorage/clusterstorage_controller_test.go
```diff
@@ -93,3 +96,114 @@ func TestSetStatusProgressing(t *testing.T) {
...
+			existingStorageClass: getFakeStorageClass(false, volumeBindingMode, "gp2", "", map[string]string{"storageclass.kubernetes.io/is-default-class": "false"}),
+			newStorageClass:      getFakeStorageClass(false, volumeBindingMode, "gp2", "", map[string]string{"storageclass.kubernetes.io/is-default-class": "true"}),
+			expectedStorageClass: getFakeStorageClass(true, volumeBindingMode, "gp2", "kubernetes.io/aws-ebs", map[string]string{"storageclass.kubernetes.io/is-default-class": "false"}),
+		},
+		{
+			name:                 "StorageClass matches and does not need to be updated",
+			existingStorageClass: getFakeStorageClass(true, volumeBindingMode, "gp2", "kubernetes.io/aws-ebs", map[string]string{"storageclass.kubernetes.io/is-default-class": "true"}),
+			newStorageClass:      getFakeStorageClass(true, volumeBindingMode, "gp2", "kubernetes.io/aws-ebs", map[string]string{"storageclass.kubernetes.io/is-default-class": "true"}),
+			expectedStorageClass: getFakeStorageClass(true, volumeBindingMode, "gp2", "kubernetes.io/aws-ebs", map[string]string{"storageclass.kubernetes.io/is-default-class": "true"}),
```
bertinatto: nit: just a suggestion to make this shorter:

We could suppress the parameters that don't need to change for know (e.g., `volumeBindingMode`, `"gp2"`) and make `getFakeStorageClass` set them as default value. Also, for the annotations we could take in only a boolean.

huffmanca: I updated the `getFakeStorageClass` to only require three arguments, and incorporated your suggestions here. After making these changes the tests continue to pass.

PR #49, opened 2019-11-20
Adding default storage class for oVirt platform

rgolangh:
Enhancment: https://github.com/openshift/enhancements/pull/61

WIP:
[] the enhancement document needs an update
[] the CSI driver needs to be published under openshift


assets/ovirt.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: thin
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"
+provisioner: kubernetes.io/ovirt-csi
```
jsafrane: `kubernetes.io` prefixes are reserved for Kubernetes and we expect that they're in-tree. You should use something else.

cuppett:
@jsafrane @gnufied Can we get this one reviewed and merged (enables higher level RHV PRs to begin to merge enabling the platform)?

assets/ovirt.yaml
gnufied: Just to be sure, this is the name of the CSI driver too right? Provisioner requires that driver name should be same as provisioner name here.

jsafrane:
i talked to @rgolangh the other day, the PR is OK, however, it has little sense without an actual CSI driver. This PR should be the last step of CSI driver installation.

PR #51, opened 2020-03-25
Bug 1816941: Move to openshift-cluster-storage-operators namespace

jsafrane:
* Merge openshift-cluster-storage-operator and openshift-cluster-storage-operators namespaces.
* Delete old operator namespace (it is not deleted by CVO)
    * Add event handler for Namespaces and remove old namespace when it exists.
    * Make sure conditions are synced correctly:
         * Available = true when storage class exists.
         * Progressing = true when the operator has some work to do:
             * Either storage class does not exist.
             * Or the old namespace exists.
             * Or both.

WIP:
 - ~test must-gather~
 - test upgrade from 4.4

### DIFF
--- a/manifests/00-namespace.yaml
+++ b/manifests/00-namespace.yaml
@@ -1,6 +1,8 @@
 apiVersion: v1
 kind: Namespace
 metadata:
+  name: openshift-cluster-storage-operators
   annotations:
     openshift.io/node-selector: ""
-  name: openshift-cluster-storage-operator
+  labels:
+    openshift.io/cluster-monitoring: "true"
--- a/manifests/01-cluster-role-binding.yaml
+++ b/manifests/01-cluster-role-binding.yaml
@@ -5,7 +5,7 @@ metadata:
 subjects:
 - kind: ServiceAccount
   name: cluster-storage-operator
-  namespace: openshift-cluster-storage-operator
+  namespace: openshift-cluster-storage-operators
 roleRef:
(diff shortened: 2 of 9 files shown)


jsafrane:
Testing upgrade from 4.4 to this PR:

1. CVO does not remove the old openshift-cluster-storage-operator namespace. Both storage operators are running, each in its own namespace.

2. CVO only updates ClusterRoleBindings, so the old operator does not have permissions to do anything and its log is full of errors (failed to update ClusterOperator, failed to watch v1.StorageClass etc.).

3. The new operator took ownership of ClusterOperator and reports upgrade finished && restores StorageClass if it does not exist.

TODO: the new operator should delete the old one + the old namespace. RBAC rules updated by CVO ensured that there is only one operator actually active (or the overlap is very short).

jsafrane:
Tested must-gather with this PR (fresh cluster installation, no upgrade): both storage-operator and csi-snapshot-controller-operator logs are present in openshift-cluster-storage-operators namespace dump.

jsafrane:
cc @openshift/storage, please review carefully

jsafrane:
/hold
maybe we can live with `openshift-cluster-storage-operator`

jsafrane:
We have chosen to use openshift-cluster-storage-operator namespace for everything.
See https://github.com/openshift/cluster-csi-snapshot-controller-operator/pull/34
/close

PR #55, opened 2020-06-01
Bug 1835869: Included check for v1alpha VolumeSnapshot CRD

huffmanca:
This includes a check for the VolumeSnapshot CRD. If we detect v1alpha1 of the CRD, then we mark the Operator as Available, but Upgradeable=false.

Note that this needs to be included in 4.3; however, I think it would be useful to have some of these validations (since we're going to be checking for CSI Drivers) in current versions as well.

### DIFF
--- a/pkg/controller/clusterstorage/clusterstorage_controller.go
+++ b/pkg/controller/clusterstorage/clusterstorage_controller.go
@@ -8,12 +8,14 @@ import (
 	"reflect"
 
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-storage-operator/pkg/controller/validation"
 	"github.com/openshift/cluster-storage-operator/pkg/generated"
 	v1helpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
 	ocontroller "github.com/openshift/library-go/pkg/controller"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
 	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
+	apiextv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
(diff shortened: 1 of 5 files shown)


sdodson:
If the detected cluster state goes away, ie: they uninstall whatever installed v1alpha volumesnapshot CRD, is Upgradeable=False unset?

pkg/controller/clusterstorage/clusterstorage_controller.go
```diff
@@ -145,6 +152,15 @@ func (r *ReconcileClusterStorage) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
+	// Run validation checks
+	err = validation.CheckAlphaSnapshot(r.client)
```
jsafrane, line 156: How often is `Reconcile` called? In other words, how long does it take for `Upgradeable=false` to get away after user removes v1alpha1 CRD?

huffmanca:
@sdodson , Yes - once the CRD is removed Upgradeable is set back to `true`. This happens fairly quickly, as we requeue the Reconcile request on an error, and the detection of the v1alpha1 CRD throws an error.

pkg/controller/clusterstorage/clusterstorage_controller.go
huffmanca: This happens fairly quickly - we requeue the request when throwing an error, and on line 161 we return an error, so it continues trying. From testing it takes seconds to clear the flag after the CRD is removed.

The other way (flagging the cluster as Upgradeable=false) takes a bit longer, as we wait for a new Reconcile request to come in and proceed through this logic. From testing it seems to be about 15 minutes.

huffmanca: Performing additional tests show that even though the Reconcile request is retried fairly frequently. In most cases it's only a couple of seconds, as the request is immediately re-queued; however, it may take the full amount of time for a subsequent Reconcile request. So up to 15 minutes in worst case scenario.

PR #57, opened 2020-06-16
Bug 1835869: Updates validation to check for VolumeSnapshotClass and VolumeSnapshotContent

huffmanca:
Continues the work in BZ1835869 to check for VolumeSnapshotClass and VolumeSnapshotContent. This puts the CRDs into a map and runs through them, building up the final error message along the way.

### DIFF
--- a/pkg/controller/clusterstorage/clusterstorage_controller.go
+++ b/pkg/controller/clusterstorage/clusterstorage_controller.go
@@ -89,6 +89,22 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for changes to second resource CustomResourceDefinition and requeue the Infrastructure
+	// This is necessary to check if VolumeSnapshot CRDs are installed after the Operator
+	err = c.Watch(&source.Kind{Type: &apiextv1beta1.CustomResourceDefinition{}}, &handler.EnqueueRequestsFromMapFunc{
+		ToRequests: handler.ToRequestsFunc(func(a handler.MapObject) []reconcile.Request {
+			return []reconcile.Request{
+				{NamespacedName: types.NamespacedName{
+					Namespace: corev1.NamespaceAll,
+					Name:      infrastructureName,
+				}},
+			}
+		}),
+	})
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -152,11 +168,11 @@ func (r *ReconcileClusterStorage) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
-	// Run validation checks
+	// Run validation checks for VolumeSnapshot, VolumeSnapshotClass, VolumeSnapshotContent
 	err = validation.CheckAlphaSnapshot(r.client)
 	if err != nil {
-		if apierrors.IsConflict(err) {
-			r.setStatusUnupgradeable(clusterOperatorInstance, "v1alpha1 VolumeSnapshot CRD found; it must be removed to upgrade")
+		if err, ok := err.(*validation.AlphaVersionError); ok {
+			r.setStatusUnupgradeable(clusterOperatorInstance, err.Error())
 		}
 		return reconcile.Result{}, err
 	}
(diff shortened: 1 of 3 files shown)


pkg/controller/validation/check_snapshot.go
```diff
@@ -16,32 +16,48 @@ import (
...
+	found := false
+	crd := &apiextv1beta1.CustomResourceDefinition{}
+	for k, v := range crdMap {
+		err := client.Get(context.TODO(), types.NamespacedName{Name: v, Namespace: corev1.NamespaceAll}, crd)
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				errMsg := "Error attempting to obtain existing " + k + " CRD."
+				log.Error(err, errMsg)
+				return err
+			}
```
jsafrane, line 44: Should you stop processing the CRD on `IsNotFound` error? (i.e. `continue` here, with a comment?)

pkg/controller/validation/check_snapshot.go
```diff
@@ -16,32 +16,48 @@ import (
...
-			err = apierrors.NewConflict(schema.GroupResource{Resource: "VolumeSnapshot"},
-				"v1alpha1 VolumeSnapshot installed.", nil)
-			log.Error(err, "Unable to update cluster as VolumeSnapshot v1alpha1 is detected. Remove this version to allow the upgrade to proceed.")
-			return err
+		for _, version := range crd.Spec.Versions {
+			if version.Name == alphaVersion {
+				found = true
+				resource = k
+				errString := "v1alpha1 " + k + " installed"
+				msgString = msgString + "Unable to update cluster as " + k + " " + alphaVersion + " is detected. Remove this version to allow the upgrade to proceed. "
```
jsafrane, line 52: If all three CRDs are installed, then the message will be really long. Maybe:
"Unable to update cluster as v1alpha1 version of volumesnapshots.snapshot.storage.k8s.io, volumesnapshotcontents.snapshot.storage.k8s.io, volumesnapshotsclasses.snapshot.storage.k8s.io is detected. Remove this version to allow the upgrade to proceed"

i.e. `strings.Join(foundCRDs, ", ")`

pkg/controller/validation/check_snapshot.go
huffmanca: I updated this to log a debug message and then `continue` to skip the rest of the loop.

pkg/controller/validation/check_snapshot.go
huffmanca: I've modified this message to build up a slice of the detected CRDs, which is used to create the final message. Since we're creating this slice, I also changed the condition from a boolean to check and see if the slice isn't empty.

Let me know your thoughts on this.

pkg/controller/clusterstorage/clusterstorage_controller.go
```diff
@@ -89,6 +89,22 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for changes to second resource CustomResourceDefinition and requeue the Infrastructure
+	// This is necessary to check if VolumeSnapshot CRDs are installed after the Operator
+	err = c.Watch(&source.Kind{Type: &apiextv1beta1.CustomResourceDefinition{}}, &handler.EnqueueRequestsFromMapFunc{
```
jsafrane, line 94: nit: Why apiextv1beta1 and not apiextv1? We will need to update to v1 at some point.

pkg/controller/validation/check_snapshot.go
```diff
@@ -16,32 +17,46 @@ import (
...
+		for _, version := range crd.Spec.Versions {
+			if version.Name == alphaVersion {
+				foundCRD = append(foundCRD, v)
+				log.Error(err, "v1alpha1 "+k+" installed")
+			}
 		}
 	}
+	if len(foundCRD) > 0 {
+		errString := "Unable to update cluster as v1alpha1 version of " + strings.Join(foundCRD, ", ") + "is detected. Remove these CRDs to allow the upgrade to proceed."
+		err := apierrors.NewConflict(schema.GroupResource{Resource: "Cluster Storage Operator"},
```
jsafrane, line 57: Why API ConflictError? That says an Update/Patch did not succeed because someone has updated the object already. I'd prefer a new error type.

huffmanca: I assumed Conflict because it said:

> the requested operation cannot be completed due to a conflict in the operation.

I thought the Conflict would be attempting to install the CRDs while a conflicting CRD (v1alpha1) was already installed. If I want to create a new error type, then I'd need to update the k8s.io/apimachinery repo? And call it `VersionConflict` ?

jsafrane: No, just use a new go error type in check_snapshot.go, you don't need to be limited to API errors.

pkg/controller/clusterstorage/clusterstorage_controller.go
huffmanca: I've had numerous errors when attempting to put this as v1. I can try to sort through these, but I think we should do the upgrade in a separate commit.

pkg/controller/validation/check_snapshot.go
huffmanca: I created a new error, `AlphaVersionError`, and updated the clusterstorage_controller to check for this type of error for validation purposes.

PR #58, opened 2020-06-18
Refactor to library-go

jsafrane:
Refactor the operator to library-go. It adds `Storage` `operator.openshift.io` CRD / CR, so the operator has its own object now.

It introduces two new library-go-style controllers:
* `DefaultStorageClassController` creates the default storage class, based on Infrastructure CR.
* `SnapshotCRDController` sets SnapshotCRDControllerUpgradeable condition to False when v1alpha1 snapshot CRDs are detected.

In addition, these standard library-go controllers are running:
* `ClusterOperatorStatusController`: Translates conditions of the above controllers (e.g. `DefaultStorageClassControllerAvailable` or `SnapshotCRDControllerUpgradeable`  to ClusterOperator conditions (`Available`, `Upgradeable`, ...)
* `OperatorManagementStateController` - manages `cr.Spec.Managed` field, reporting `ManagementStateDegraded` condition when it has a wrong value.
* `ClusterOperatorLoggingController` - manages klog log level based on `cr.Spec.OperatorLogLevel`.

And move to go mod (with huge vendor changes).


cmd/cluster-storage-operator/main.go
```diff
@@ -0,0 +1,55 @@
...
+		},
+	}
+
+	ctrlCmd := controllercmd.NewControllerCommandConfig(
+		"cluster-storage-operator",
+		version.Get(),
+		operator.RunOperator,
+	).NewCommand()
+	ctrlCmd.Use = "start"
+	ctrlCmd.Short = "Start the ACluster Storage Operator"
```
gnufied, line 50: typo?

gnufied:
@jsafrane this https://github.com/openshift/cluster-storage-operator/pull/58/commits/5ab42422d9e194ff9a5d6322e3ec06925b525c6b commit is so big that it does not even load properly in the browser and everything becomes very sluggish. I know it can be bit pain but can we split that commit in two? One that actually contains our code and then generated code in separate commit

manifests/01-role-binding-authentication-reader.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: foo-role
+  namespace: kube-system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: extension-apiserver-authentication-reader
+subjects:
+- kind: ServiceAccount
+  name: cluster-storage-operator
+  namespace: openshift-cluster-storage-operator
```
gnufied, line 14: Do we need rbac permissions to read/update `storage` object CR? Also - I must have missed something but I can't find a place where `cluster` object is created.

jsafrane: good catch, I forgot to push the last commit. It's created by CVO now.

cmd/cluster-storage-operator/main.go
jsafrane: fixed

jsafrane:
/test images
That one is flake.

Looking at e2e-aws must-gather errors.

jsafrane:
/retest

Must-gather failed because:
```
Expected
    <string>: [...]/namespaces/openshift-kube-apiserver/core/configmaps.yaml
to exist
```
That should not be affected by any change in cluster-storage-operator.

pkg/operator/starter.go
```diff
@@ -0,0 +1,146 @@
...
+	cfgInformers := cfginformers.NewSharedInformerFactoryWithOptions(cfgClientset, resync)
+
+	apiExtClientset, err := apiextclient.NewForConfig(controllerConfig.KubeConfig)
+	if err != nil {
+		return err
+	}
+	apiExtInformers := apiextinformers.NewSharedInformerFactoryWithOptions(apiExtClientset, resync)
+
+	operatorClient := &operatorclient.OperatorClient{
+		Informers: operatorInformers,
```
gnufied, line 71: So we are adding 4 differnet cliensets and it appears that `NewForConfig` itself does not document intended purpose. Can we document at each initialization point why we are initializing 4 different cliensets?

pkg/operator/defaultstorageclass/controller.go
```diff
@@ -0,0 +1,182 @@
...
+		kubeClient:         kubeClient,
+		infraLister:        infraInformer.Config().V1().Infrastructures().Lister(),
+		storageClassLister: kubeInformers.Storage().V1().StorageClasses().Lister(),
+		eventRecorder:      eventRecorder,
+	}
+	return factory.New().WithSync(c.sync).WithSyncDegradedOnError(operatorClient).WithInformers(
+		operatorClient.Informer(),
+		infraInformer.Config().V1().Infrastructures().Informer(),
+		kubeInformers.Storage().V1().StorageClasses().Informer(),
+	).ToController("DefaultStorageClassController", eventRecorder)
```
gnufied, line 67: Do we need to also call `ResyncEver` here? From docs:

>// ResyncEvery will cause the Sync() function to be called periodically, regardless of informers.

pkg/operator/snapshotcrd/controller.go
```diff
@@ -0,0 +1,124 @@
...
+	if len(alphaCRDs) > 0 {
+		msg := fmt.Sprintf("Unable to update cluster as v1alpha1 version of %s is detected. Remove these CRDs to allow the upgrade to proceed.", strings.Join(alphaCRDs, ", "))
+		upgradeable.Status = operatorapi.ConditionFalse
+		upgradeable.Message = msg
+		upgradeable.Reason = "AlphaDetected"
+	}
+
+	if _, _, updateErr := v1helpers.UpdateStatus(c.operatorClient,
+		v1helpers.UpdateConditionFn(upgradeable),
+	); updateErr != nil {
```
gnufied, line 89: If I am reading this code right - will this result in existing conditions being replaced with new condition? We are also missing test for that code path. But may be this belongs to different PR.

pkg/operator/snapshotcrd/controller_test.go
```diff
@@ -0,0 +1,277 @@
...
+					getCRD("volumesnapshotclassess.snapshot.storage.k8s.io", "v1alpha1"),
+				},
+			},
+			expectedObjects: testObjects{
+				storage: getCR(
+					withFalseConditions(conditionsPrefix + opv1.OperatorStatusTypeUpgradeable),
+				),
+			},
+			expectErr: false,
+		},
```
gnufied, line 222: Can we add test for case where there were other conditions on the CR and they should not be replaced when Upgradable condition is changed?

pkg/operator/defaultstorageclass/controller.go
jsafrane: Each informer already has 20 minutes resync interval.

pkg/operator/snapshotcrd/controller.go
jsafrane: Existing conditions will be kept, will add unit test as suggested below.

pkg/operator/snapshotcrd/controller_test.go
jsafrane: Test added

pkg/operator/starter.go
jsafrane: Added a comment for each client

PR #62, opened 2020-07-20
Fix installation on usupported platforms like oVirt

jsafrane:
The operator should mark itself as `Available` when on a platform that does not have a default StorageClass (yet).

library-go status syncer expects at least Available condition in `storage` CR to mark the OperatorStatus as `Available`.

cc @openshift/storage

### DIFF
--- a/pkg/operator/defaultstorageclass/controller.go
+++ b/pkg/operator/defaultstorageclass/controller.go
@@ -98,10 +98,17 @@ func (c *Controller) sync(ctx context.Context, syncCtx factory.SyncContext) erro
 				Reason:  "UnsupportedPlatform",
 				Message: syncErr.Error(),
 			}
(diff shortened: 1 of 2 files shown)


pkg/operator/defaultstorageclass/controller.go
```diff
@@ -98,10 +98,13 @@ func (c *Controller) sync(ctx context.Context, syncCtx factory.SyncContext) erro
 				Reason:  "UnsupportedPlatform",
 				Message: syncErr.Error(),
 			}
+			availableCnd.Message = "No default StorageClass for this platform"
```
bertinatto, line 101: Could you add a brief comment saying why we set it to true (like the above)?

After reading the message, one may assume the condition is wrong

jsafrane: A comment added, PTAL

PR #65, opened 2020-07-29
Add oVirt and manila to CSO

jsafrane:
* oVirt is simple
* Manila is an optional CSI driver.
  * If underlying cloud has Manila, everything is the same as for the other (mandatory) drivers
  * If there is no Manila, Manila operator reports condition `Disabled: true`. CSO then consumes it and reports `Available: true` (because there is nothing wrong in the cluster). User can get details about Manila in `Storage` CR.

@openshift/storage 

/hold
requires https://github.com/openshift/csi-driver-manila-operator/pull/44 merged first

### DIFF
--- a/assets/csidriveroperators/ovirt/07_deployment.yaml
+++ b/assets/csidriveroperators/ovirt/07_deployment.yaml
@@ -1,0 +1,120 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: ovirt-csi-driver-operator
+  namespace: openshift-cluster-csi-drivers
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      name: ovirt-csi-driver-operator
+  template:
+    metadata:
+      labels:
+        name: ovirt-csi-driver-operator
+    spec:
+      serviceAccountName: ovirt-csi-driver-operator
+      priorityClassName: system-cluster-critical
+      initContainers:
+        - name: prepare-ovirt-config
+          env:
+            - name: OVIRT_URL
+              valueFrom:
+                secretKeyRef:
+                  name: ovirt-credentials
+                  key: ovirt_url
+            - name: OVIRT_USERNAME
+              valueFrom:
+                secretKeyRef:
+                  name: ovirt-credentials
+                  key: ovirt_username
+            - name: OVIRT_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: ovirt-credentials
+                  key: ovirt_password
+            - name: OVIRT_CAFILE
+              value: /tmp/config/ovirt-engine-ca.pem
+            - name: OVIRT_INSECURE
+              valueFrom:
+                secretKeyRef:
+                  name: ovirt-credentials
+                  key: ovirt_insecure
+            - name: OVIRT_CA_BUNDLE
+              valueFrom:
+                secretKeyRef:
+                  name: ovirt-credentials
+                  key: ovirt_ca_bundle
+          image: ${OPERATOR_IMAGE}
+          resources:
+            requests:
+              memory: 50Mi
+              cpu: 10m
+          command:
+            - /bin/sh
+            - -c
+            - |
+              #!/bin/sh
+              cat << EOF > /tmp/config/ovirt-config.yaml
+              ovirt_url: $OVIRT_URL
+              ovirt_username: $OVIRT_USERNAME
+              ovirt_password: $OVIRT_PASSWORD
+              # set a valid path only if ca bundle has content
+              ovirt_cafile: ${OVIRT_CA_BUNDLE:+$OVIRT_CAFILE}
+              ovirt_insecure: $OVIRT_INSECURE
+              EOF
+              if [[ -n "$OVIRT_CA_BUNDLE" ]]; then echo "$OVIRT_CA_BUNDLE" > $OVIRT_CAFILE ; fi
+          volumeMounts:
+            - name: config
+              mountPath: /tmp/config
+
+      containers:
+        - name: ovirt-csi-driver-operator
+          image: ${OPERATOR_IMAGE}
+          imagePullPolicy: IfNotPresent
+          resources:
+            requests:
+              memory: 50Mi
+              cpu: 10m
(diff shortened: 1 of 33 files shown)


assets/csidriveroperators/ovirt/07_deployment.yaml
```diff
@@ -0,0 +1,52 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      name: ovirt-csi-driver-operator
+  template:
+    metadata:
+      labels:
+        name: ovirt-csi-driver-operator
+    spec:
+      serviceAccountName: ovirt-csi-driver-operator
```
rgolangh, line 17: 07_deployment.yaml was just updated, please pull it again.

jsafrane: Updated. And added CredentialRequest sync for oVirt to CSO.

assets/csidriveroperators/ovirt/07_deployment.yaml
```diff
@@ -0,0 +1,119 @@
...
+          resources:
+            requests:
+              memory: 50Mi
+              cpu: 10m
+          tolerations:
+            - key: CriticalAddonsOnly
+              operator: Exists
+          args:
+            - start
+            - "--node=$(KUBE_NODE_NAME)"
```
bertinatto, line 85: What about adding ${LOG_LEVEL}?

jsafrane: Fixed

pkg/operator/csidriveroperator/crcontroller.go
```diff
@@ -240,6 +221,61 @@ func (c *CSIDriverOperatorCRController) applyClusterCSIDriver(requiredOriginal *
...
+			progressingCnd.Reason = "WaitForOperator"
+			progressingCnd.Message = fmt.Sprintf("Waiting for %s operator to report status", c.name)
+		}
+	}
+
+	degradedCnd := status.UnionCondition(operatorapi.OperatorStatusTypeDegraded, operatorapi.ConditionFalse, nil, conditions...)
+	degradedCnd.Type = c.crConditionName(operatorapi.OperatorStatusTypeDegraded)
+	if degradedCnd.Status == operatorapi.ConditionUnknown {
+		degradedCnd.Status = operatorapi.ConditionFalse
+	}
```
bertinatto, line 259: Can we have an operator that's disabled&optional but is also progressing or degraded?

Doesn't sound like we want to handle that, but just checking...

pkg/operator/csidriveroperator/csioperatorclient/ovirt.go
```diff
@@ -0,0 +1,58 @@
...
+			"csidriveroperators/ovirt/03_role.yaml",
+			"csidriveroperators/ovirt/04_rolebinding.yaml",
+			"csidriveroperators/ovirt/05_clusterrole.yaml",
+			"csidriveroperators/ovirt/06_clusterrolebinding.yaml",
+		},
+		CRAsset:         "csidriveroperators/ovirt/08_cr.yaml",
+		DeploymentAsset: "csidriveroperators/ovirt/07_deployment.yaml",
+		ImageReplacer:   strings.NewReplacer(pairs...),
+		ExtraControllers: []factory.Controller{
+			newOVirtCredentialsRequest(clients, recorder),
```
bertinatto, line 43: Why are we creating the credentials request in CSO for oVirt?

oVirt operator creates it as well: https://github.com/openshift/ovirt-csi-driver-operator/blob/master/pkg/operator/starter.go#L86

jsafrane: oVirt operator needs to check name of storage domain from oVirt to create the default storage class.

bertinatto: Ack.

@rgolangh does that mean the oVirt operator doesn't need to create the credentialsrequestcontroller [1]?

[1] https://github.com/openshift/ovirt-csi-driver-operator/blob/master/pkg/operator/starter.go#L86

pkg/operator/csidriveroperator/crcontroller.go
jsafrane: Disabled & degraded can happen:

1. Manila operator checks there is no Manila in OpenStack and sets `Disabled: true`
2. On regular re-sync, Manila fails to talk to OpenStack and reports `Degraded: true`. This keeps `Disabled: true` (at least I think so)

With the current code, CSO will report `Degraded: true`.

Disabled & Progressing is IMO impossible, but CSO should be robust enough to handle that.

pkg/operator/csidriveroperator/csioperatorclient/ovirt.go
jsafrane: Maybe we could fix oVirt operator to reuse secret provided by CSO's CredentialRequest.

pkg/operator/csidriveroperator/crcontroller.go
```diff
@@ -240,6 +221,61 @@ func (c *CSIDriverOperatorCRController) applyClusterCSIDriver(requiredOriginal *
 	return actual, true, err
 }
 
+func (c *CSIDriverOperatorCRController) syncConditions(conditions []operatorapi.OperatorCondition, updatefn v1helpers.UpdateStatusFunc) error {
```
bertinatto, line 224: Do you think it would be more readable to split this into 2 functions: one to handle optional operators and the other to handle non-optional ones? Or have a big if-else clause (if c.optional)?

jsafrane: I tried and it looks even worse - two functions that look like copies, with only small differences.

jsafrane:
The last ovirt tests almost succeeded. Only this one failed:

`[sig-instrumentation] Prometheus when installed on the cluster shouldn't report any alerts in firing state apart from Watchdog and AlertmanagerReceiversNotConfigured [Early] [Suite:openshift/conformance/parallel] `. And the alert do not look storage related (memory overcommit on nodes)

PR #66, opened 2020-07-30
Add migration from OLM CSI driver operators

jsafrane:
OLMOperatorRemovalController moves CSI driver operator from OLM to CSO. It must run before CSO starts CSI driver operator to make sure only one  operator / only one driver runs at a time.

~It needs #65 merged first.~

### DIFF
--- a/pkg/operator/csidriveroperator/driverstarter.go
+++ b/pkg/operator/csidriveroperator/driverstarter.go
@@ -10,13 +10,13 @@ import (
 	"github.com/openshift/cluster-storage-operator/pkg/csoclients"
 	"github.com/openshift/cluster-storage-operator/pkg/generated"
 	"github.com/openshift/cluster-storage-operator/pkg/operator/csidriveroperator/csioperatorclient"
+	"github.com/openshift/cluster-storage-operator/pkg/operatorclient"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/controller/manager"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/staticresourcecontroller"
 	"github.com/openshift/library-go/pkg/operator/status"
-	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	"k8s.io/klog/v2"
 )
 
@@ -25,12 +25,12 @@ const (
 )
 
 // This CSIDriverStarterController starts CSI driver controllers based on the
-// underlying cloud. It does not install anything by itself, only monitors
-// Infrastructure instance and starts individual ControllerManagers for the
-// particular cloud. It produces following Conditions:
+// underlying cloud and removes it from OLM. It does not install anything by
+// itself, only monitors Infrastructure instance and starts individual
+// ControllerManagers for the particular cloud. It produces following Conditions:
 // CSIDriverStarterDegraded - error checking the Infrastructure
 type CSIDriverStarterController struct {
-	operatorClient v1helpers.OperatorClient
+	operatorClient *operatorclient.OperatorClient
 	infraLister    openshiftv1.InfrastructureLister
 	versionGetter  status.VersionGetter
 	targetVersion  string
@@ -41,8 +41,14 @@ type CSIDriverStarterController struct {
 
 type csiDriverControllerManager struct {
 	operatorConfig csioperatorclient.CSIOperatorConfig
-	mgr            manager.ControllerManager
-	running        bool
+	// Controller that deletes old OLM-based operator (optional).
+	olmRemovalController        *OLMOperatorRemovalController
+	olmRemovalControllerRunning bool
+	// ControllerManager that installs the CSI driver operator and all its
+	// objects. It is started after olmRemovalController confirms the old
+	// driver is gone.
+	driverManager        manager.ControllerManager
+	driverManagerRunning bool
 }
 
 func NewCSIDriverStarterController(
@@ -69,9 +75,11 @@ func NewCSIDriverStarterController(
 	c.controllers = []csiDriverControllerManager{}
 	for _, cfg := range driverConfigs {
 		c.controllers = append(c.controllers, csiDriverControllerManager{
-			operatorConfig: cfg,
-			mgr:            c.createCSIControllerManager(cfg, clients, resyncInterval),
-			running:        false,
+			operatorConfig:              cfg,
+			olmRemovalController:        NewOLMOperatorRemovalController(cfg, clients, c.eventRecorder, resyncInterval),
+			olmRemovalControllerRunning: false,
+			driverManager:               c.createCSIControllerManager(cfg, clients, resyncInterval),
+			driverManagerRunning:        false,
 		})
 	}
 
@@ -82,6 +90,9 @@ func NewCSIDriverStarterController(
 }
 
 func (c *CSIDriverStarterController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	klog.V(4).Infof("CSIDriverStarterController.Sync started")
+	defer klog.V(4).Infof("CSIDriverStarterController.Sync finished")
+
 	opSpec, _, _, err := c.operatorClient.GetOperatorState()
 	if err != nil {
 		return err
@@ -102,10 +113,35 @@ func (c *CSIDriverStarterController) sync(ctx context.Context, syncCtx factory.S
 	}
 
 	for i := range c.controllers {
-		if c.controllers[i].operatorConfig.Platform == platform && !c.controllers[i].running {
-			klog.V(2).Infof("Starting CSI driver controller manager %s", c.controllers[i].operatorConfig.ConditionPrefix)
-			go c.controllers[i].mgr.Start(ctx)
-			c.controllers[i].running = true
+		ctrl := &c.controllers[i]
+		if ctrl.operatorConfig.Platform != platform {
+			continue
+		}
+		klog.V(4).Infof("CSIDriverStarterController syncing %s", ctrl.operatorConfig.ConditionPrefix)
+
+		if ctrl.olmRemovalController != nil {
+			if !ctrl.olmRemovalControllerRunning {
+				// Start the OLM removal controller
+				klog.V(2).Infof("Starting OLM removal controller for %s", ctrl.operatorConfig.ConditionPrefix)
+				queue := syncCtx.Queue()
+				key := syncCtx.QueueKey()
+				ctrl.olmRemovalController.SetFinishedHandler(func() {
+					// Resync this controller when the migration is completed
+					queue.Add(key)
+				})
+				go ctrl.olmRemovalController.Run(ctx, 1)
+				ctrl.olmRemovalControllerRunning = true
+			}
+			// Wait until the removal finishes
+			if !ctrl.olmRemovalController.OldCSIDriverRemoved() {
+				return nil
+			}
+		}
+		// Removal is either complete or is not needed
+		if !ctrl.driverManagerRunning {
+			klog.V(2).Infof("Starting ControllerManager for %s", ctrl.operatorConfig.ConditionPrefix)
+			go ctrl.driverManager.Start(ctx)
+			ctrl.driverManagerRunning = true
 		}
 	}
 	return nil
@@ -125,13 +161,14 @@ func (c *CSIDriverStarterController) createCSIControllerManager(
 		c.operatorClient,
 		c.eventRecorder).AddKubeInformers(clients.KubeInformers), 1)
 
-	manager = manager.WithController(NewCSIDriverOperatorCRController(
+	crController := NewCSIDriverOperatorCRController(
 		cfg.ConditionPrefix,
 		clients,
 		cfg,
 		c.eventRecorder,
 		resyncInterval,
-	), 1)
+	)
+	manager = manager.WithController(crController, 1)
 
 	manager = manager.WithController(NewCSIDriverOperatorDeploymentController(
 		cfg.ConditionPrefix,
--- a/pkg/operator/csidriveroperator/driverstarter.go
+++ b/pkg/operator/csidriveroperator/driverstarter.go
@@ -41,12 +41,8 @@ type CSIDriverStarterController struct {
 
 type csiDriverControllerManager struct {
 	operatorConfig csioperatorclient.CSIOperatorConfig
-	// Controller that deletes old OLM-based operator (optional).
-	olmRemovalController        *OLMOperatorRemovalController
-	olmRemovalControllerRunning bool
 	// ControllerManager that installs the CSI driver operator and all its
-	// objects. It is started after olmRemovalController confirms the old
-	// driver is gone.
+	// objects.
 	driverManager        manager.ControllerManager
 	driverManagerRunning bool
 }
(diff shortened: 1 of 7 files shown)


pkg/operator/csidriveroperator/driverstarter.go
```diff
@@ -102,10 +113,35 @@ func (c *CSIDriverStarterController) sync(ctx context.Context, syncCtx factory.S
...
+		ctrl := &c.controllers[i]
+		if ctrl.operatorConfig.Platform != platform {
+			continue
+		}
+		klog.V(4).Infof("CSIDriverStarterController syncing %s", ctrl.operatorConfig.ConditionPrefix)
+
+		if ctrl.olmRemovalController != nil {
+			if !ctrl.olmRemovalControllerRunning {
+				// Start the OLM removal controller
+				klog.V(2).Infof("Starting OLM removal controller for %s", ctrl.operatorConfig.ConditionPrefix)
```
bertinatto, line 125: nit: v(4)?

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,478 @@
...
+		removed, err := c.deleteCSV(ctx, subNamespace, csvName)
+		if err != nil {
+			return err
+		}
+		if !removed {
+			klog.V(4).Infof("OLMOperatorRemovalController.Sync waiting for OLM CSV to disappear")
+			return c.markProgressing("Waiting for OLM CSV to be deleted")
+		}
+
+		removed, err = c.deleteCSV(ctx, subNamespace, csvName)
```
gnufied, line 148: Did we intend to this twice?

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,479 @@
...
+		if err != nil {
+			return err
+		}
+		if !removed {
+			klog.V(4).Infof("OLMOperatorRemovalController.Sync waiting for OLM CSV to disappear")
+			return c.markProgressing("Waiting for OLM CSV to be deleted")
+		}
+
+		removed, err = c.deleteSubscription(ctx, subNamespace, subName)
```
gnufied, line 157: IMO we should delete the subscription first because otherwise we are racing with OLM operator and it might try to install CSV again if it was deleted, but then we may run into a problem that if we deleted the subscription we wouldn't be able to find the current CSV?

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,463 @@
...
+	if err != nil {
+		return err
+	}
+	if !removed {
+		klog.V(4).Infof("CSIDriverMigrationController.Sync waiting for the driver Deployment to disappear")
+		return c.markProgressing("Waiting for OLM driver Deployment to be deleted")
+	}
+
+	// 5. Delete the driver DaemonSet.
+	removed, err = c.ensureDriverDaemonSetRemoved(ctx, c.olmOptions.CSIDriverNamespace, c.olmOptions.CSIDriverDaemonSetName)
```
gnufied, line 199: A pod that uses any of these PVs if gets evicted during upgrade, then it will be stuck basically?

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,463 @@
...
+	c := &OLMOperatorRemovalController{
+		name:             csiOperatorConfig.ConditionPrefix,
+		operatorClient:   clients.OperatorClient,
+		olmOptions:       csiOperatorConfig.OLMOptions,
+		dynamicClient:    clients.DynamicClient,
+		kubeClient:       clients.KubeClient,
+		nsLister:         clients.KubeInformers.InformersFor("").Core().V1().Namespaces().Lister(),
+		deploymentLister: clients.KubeInformers.InformersFor("").Apps().V1().Deployments().Lister(),
+		daemonSetLister:  clients.KubeInformers.InformersFor("").Apps().V1().DaemonSets().Lister(),
+		eventRecorder:    eventRecorder,
```
bertinatto, line 103: prefix

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,463 @@
...
+// Steps performed:
+// 1. Remove old OLM Subscription and CSV. Remember the operator namespace
+//    (in Storage CR annotation), just in case the controller is restarted
+//    after Subscription removal.
+// 2. Remove the CSI driver Deployment and DaemonSet and wait until they're
+//    really removed. Keep the old namespace around, as it can contain
+//    Manila secrets that are used by existing PVs!
+// 3. Remove the old CR (incl. removing of all of its finalizers).
+// It produces following conditions:
+// <CSI driver name>OLMOperatorRemovalProgressing/Degraded: for status reporting.
```
bertinatto, line 37: I think this controller is not reporting degraded currently

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,463 @@
...
+		if err != nil {
+			return "", "", "", false, err
+		}
+
+		if pkg == c.olmOptions.OLMPackageName && source == olmSource && sourceNamespace == olmSourceNamespace {
+			csvName, _, err := unstructured.NestedString(obj.Object, "status", "currentCSV")
+			if err != nil {
+				return "", "", "", false, err
+			}
+			klog.V(2).Infof("Found subscription %s/%s with CSV %s", obj.GetNamespace(), obj.GetName(), csvName)
```
bertinatto, line 255: V(2) might be too noisy

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,463 @@
...
+func (c *OLMOperatorRemovalController) deleteCSV(ctx context.Context, namespace, name string) (bool, error) {
+	err := c.dynamicClient.Resource(csvResourceGVR).Namespace(namespace).Delete(ctx, name, metav1.DeleteOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	}
+	klog.V(4).Infof("Deleted CSV %s/%s", namespace, name)
+	return false, nil
```
bertinatto, line 283: shouldn't you return true here?

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,463 @@
...
+func (c *OLMOperatorRemovalController) loadOperatorNamespace() (string, error) {
+	meta, err := c.operatorClient.GetObjectMeta()
+	if err != nil {
+		return "", err
+	}
+	ns := meta.Annotations[c.name+olmOperatorNamespaceAnnotation]
+	klog.V(4).Infof("Loaded operator namespace annotation %s", ns)
+	return ns, nil
+
+}
```
bertinatto, line 300: nit: empty line

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,463 @@
...
+
+func (c *OLMOperatorRemovalController) markProgressing(message string) error {
+	progressing := operatorapi.OperatorCondition{
+		Type:    c.Name() + operatorapi.OperatorStatusTypeProgressing,
+		Reason:  "RemovingFromOLM",
```
bertinatto, line 305: RemovingOLMOperator?

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,463 @@
...
+	_, err := c.daemonSetLister.DaemonSets(namespace).Get(name)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return false, err
+		}
+		return true, nil
+	}
+	err = c.kubeClient.AppsV1().DaemonSets(namespace).Delete(ctx, name, metav1.DeleteOptions{})
+	if err != nil {
+		return false, err
```
bertinatto, line 391: check for IsNotFound?

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,463 @@
...
+	_, err := c.deploymentLister.Deployments(namespace).Get(name)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return false, err
+		}
+		return true, nil
+	}
+	err = c.kubeClient.AppsV1().Deployments(namespace).Delete(ctx, name, metav1.DeleteOptions{})
+	if err != nil {
+		return false, err
```
bertinatto, line 374: check IsNotFound?

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,463 @@
...
+		if err != nil {
+			return false, err
+		}
+		klog.V(4).Infof("Deleted old CR finalizers")
+	}
+	err = c.dynamicClient.Resource(res).Delete(ctx, oldCRName, metav1.DeleteOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return true, nil
+		}
```
bertinatto, line 418: nit: some of these checks are `!apierrors.IsNotFound`, others are `apierrors.IsNotFound`. It'd be nice to use have only 1 way of checking these, it's easier to follow as I don't have to re-process if the returns are correct :)

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+	klog.V(4).Infof("Deleted driver DaemonSet %s/%s", namespace, name)
+	// Don't report the DaemonSet is removed, wait until IsNotFound error above
+	return false, nil
+}
+
+func (c *OLMOperatorRemovalController) ensureCRRemoved(ctx context.Context, res schema.GroupVersionResource) (bool, error) {
+	cr, err := c.dynamicClient.Resource(res).Get(ctx, oldCRName, metav1.GetOptions{})
+	if err != nil {
+		if !errors.IsNotFound(err) {
```
bertinatto, line 401: `apierrors` and `errors`: you might be importing the same package twice

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,463 @@
...
+	err = c.dynamicClient.Resource(res).Delete(ctx, oldCRName, metav1.DeleteOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	}
+	klog.V(4).Infof("Deleted old CR")
+	// Don't report the CR is removed, wait until IsNotFound error above
+	return false, nil
```
bertinatto, line 423: If `err == nil` can't we assume the CR was deleted correctly? This would speed things up, as we wouldn't have to wait for a resync

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,463 @@
...
+	olmSourceNamespace             = "openshift-marketplace"
+
+	oldCRName = "cluster"
+)
+
+func NewOLMOperatorRemovalController(
+	csiOperatorConfig csioperatorclient.CSIOperatorConfig,
+	clients *csoclients.Clients,
+	eventRecorder events.Recorder,
+	resyncInterval time.Duration,
```
bertinatto, line 68: Should this controller have its own resync interval?

pkg/operator/csidriveroperator/olmremovalcontroller.go
jsafrane: stuck until the new driver starts, yes

pkg/operator/csidriveroperator/driverstarter.go
jsafrane: It should be printed only once (`!ctrl.olmRemovalControllerRunning`) and I'd prefer to know what has been started.

pkg/operator/csidriveroperator/olmremovalcontroller.go
jsafrane: Not sure why. Do you think it should resync more often?

pkg/operator/csidriveroperator/olmremovalcontroller.go
jsafrane: I'd rather have it confirmed from the API server...

pkg/operator/csidriveroperator/olmremovalcontroller.go
jsafrane: good catch! using `apierrors`.

pkg/operator/csidriveroperator/olmremovalcontroller.go
jsafrane: Unified

pkg/operator/csidriveroperator/olmremovalcontroller.go
jsafrane: added

pkg/operator/csidriveroperator/olmremovalcontroller.go
jsafrane: Added

pkg/operator/csidriveroperator/olmremovalcontroller.go
jsafrane: Fixed

pkg/operator/csidriveroperator/olmremovalcontroller.go
jsafrane: No, `Delete()` success does not mean the object is deleted. I added a comment there.

pkg/operator/csidriveroperator/olmremovalcontroller.go
jsafrane: Moved to v(4), however, it should be printed just once in ideal case.

pkg/operator/csidriveroperator/olmremovalcontroller.go
jsafrane: Added

pkg/operator/csidriveroperator/olmremovalcontroller.go
jsafrane: Fixed :-)

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,463 @@
...
+	cr, err := c.dynamicClient.Resource(res).Get(ctx, oldCRName, metav1.GetOptions{})
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return false, err
+		}
+		return true, nil
+	}
+	if len(cr.GetFinalizers()) > 0 {
+		cr.SetFinalizers([]string{})
+		cr, err = c.dynamicClient.Resource(res).Update(ctx, cr, metav1.UpdateOptions{})
```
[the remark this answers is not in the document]
jsafrane: Fixed

pkg/operator/csidriveroperator/olmremovalcontroller.go
jsafrane: No! Fixed.

pkg/operator/csidriveroperator/olmremovalcontroller.go
bertinatto: Say an error happens during sync, the controller will report a degraded condition and wait 20 (for example) minutes to resync; in the meantime, the DeploymentController will wait to deploy the CSI Driver operator...

If this is correct, IMO we should special-case this controller and hardcode a faster (but not much) resync interval... does this makes sense?

bertinatto: BTW, this can be important if we'll rely on resyncs like here: https://github.com/openshift/cluster-storage-operator/pull/66#discussion_r463617363

jsafrane:
/close
We decided not to move the operand namespace. New PR will follow.

PR #69, opened 2020-08-03
Bug 1862523: Add migration controller

jsafrane:
OLMOperatorRemovalController removes CSI driver operator from OLM. It must run before CSO starts CSI driver operator to make sure only one operator / only one driver runs at a time.

It removes Subscription, CSV, waits for the old operator Deployment to get removed by OLM and finally it force-removes the old CR (by clearing all its finalizers).

It signalizes that it's finished with OLM removal by setting `ManilaOLMOperatorRemovalAvailable: true`. This is then recognized by CSIDriverStarterController that runs the new operator.

### DIFF
--- a/pkg/operator/csidriveroperator/olmremovalcontroller.go
+++ b/pkg/operator/csidriveroperator/olmremovalcontroller.go
@@ -1,0 +1,409 @@
+package csidriveroperator
+
+import (
+	"context"
+	"time"
+
+	operatorapi "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-storage-operator/pkg/csoclients"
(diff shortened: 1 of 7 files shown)


jsafrane:
cc @openshift/storage

pkg/operator/csidriveroperator/olmremovalcontroller.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+	if opSpec.ManagementState != operatorapi.Managed {
+		return nil
+	}
+
+	// 1. Find subscription + namespace
+	subNamespace, subName, csvName, found, err := c.findSubscription(ctx)
+	if err != nil {
+		return err
+	}
```
bertinatto, line 119: Is there a way to short-circuit here in case everything has been deleted already?

PR #71, opened 2020-08-05
Bug 1865769: Remove authentication rolebinding

jsafrane:
CSO is cluster-admin and can read anything, therefore it does not need an extra role to read stuff in kube-system namespace

@openshift/storage


jsafrane:
I reordered the manifests

PR #73, opened 2020-08-10
Bug 1867704: Give pod listing permissions to aws operator

gnufied:
Please note this is a temporary measure, I intend to remove these permissions and disable the support for volume-in-use error in resizer for EBS because it is not applicable.

However, without this fix - we have a chicken and egg problem. The rebase PR of external-resizer will not merge until external-resizer can list the pods.


jsafrane:
/lgtm
Please update https://github.com/openshift/aws-ebs-csi-driver-operator/tree/master/manifests too

PR #75, opened 2020-08-13
Bug 1867735: Fix syncing of Manila CA certificate

jsafrane:
The sync failed, because CSO's KubeInformersForNamespaces did not watch openshift-config namespace.

Add the namespace to the namespace list + make sure that CSO panics on similar errors in future.


jsafrane:
I can see `storage` is `Available: true` in the openstack CI job.
/retest

PR #76, opened 2020-08-18
Bug 1866262: Remove DriverConfig from CSI operator CRs

jsafrane:
We do not have use for the field, so remove it. Bump openshift/api to get the updated type.


jsafrane:
cc @openshift/storage

bertinatto:
Will you update the manifests from CSI operators repos as well?

PR #77, opened 2020-08-18
Bug 1869516: Prevent crash if PlatformStatus is nil

huffmanca:
There was an issue in [BZ1869516](https://bugzilla.redhat.com/show_bug.cgi?id=1869516) where the CSO paniced during an upgrade on a baremetal UPI cluster. When examining the Infrastructure resource, there was no PlatformStatus defined.

This should check and return an unsupportedPlatform if the PlatformStatus is nil.

### DIFF
--- a/pkg/operator/defaultstorageclass/controller.go
+++ b/pkg/operator/defaultstorageclass/controller.go
@@ -134,6 +134,11 @@ func (c *Controller) syncStorageClass() error {
 	if err != nil {
 		return err
 	}
+	// Check to see if the PlatformStatus is nil. This has been seen on some
+	// UPI installs on baremetal platforms
+	if infrastructure.Status.PlatformStatus == nil {
+		return unsupportedPlatformError
+	}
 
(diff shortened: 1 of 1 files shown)


pkg/operator/defaultstorageclass/controller.go
```diff
@@ -160,6 +160,9 @@ func (c *Controller) syncStorageClass() error {
 }
 
 func newStorageClassForCluster(infrastructure *configv1.Infrastructure) (*storagev1.StorageClass, error) {
+	if infrastructure.Status.PlatformStatus == nil {
+		return nil, unsupportedPlatformError
+	}
```
bertinatto, line 165: Good catch!

Just a nit: I'd suggest moving this check right after `infrastructure` is fetched.

huffmanca: I moved this up to `syncStorageClass()` and included a comment about why we're performing this check.

bertinatto (commented):
/lgtm

PR #78, opened 2020-08-21
Bug 1871051: Add CSI operators permissions to patch volumeattachment/status

jsafrane:
The external-attacher needs to patch volumeattachment/status. Add such permission to AWS EBS and oVirt operators, so they're able to grant it to the driver.

@openshift/storage


deads2k:
/override ci/prow/e2e-aws

failed on https://bugzilla.redhat.com/show_bug.cgi?id=1871987 but otherwise worked

PR #80, opened 2020-08-27
Bug 1868339: Move CredentialsRequest creation to CSO

bertinatto:
CSI operators should let CSO apply the CredentialsRequest objects, regardless of the cloud platform.

This patch is a requirement to allow users to manually see the secrets they need to create when running cloud-credentials-controller in Manual mode. More information at: https://github.com/openshift/cloud-credential-operator/blob/master/docs/mode-manual-creds.md.

Once this is merged, CredentialsRequests will be created by CVO, so we'll need follow up PRs to CSI operator repositories to **not create those objects themselves**.

Please see individual commits for a description of each change.

CC @openshift/storage


bertinatto:
/hold

Before this merges I need to patch the CSI driver operators and then test manually.

bertinatto:
Also, @dgoodwin FYI

dgoodwin:
Thank you for this, looks good from my PoV.

bertinatto:
/test e2e-aws

`Operator degraded (IngressControllersDegraded): Some ingresscontrollers are degraded: default`

tsmetana (approved):
/lgtm

gnufied:
> Once this is merged, CredentialsRequests will be created by CVO, so we'll need follow up PRs to CSI operator repositories to not create those objects themselves.

@bertinatto Can CCO handle cases where secret was already created? Looks like it already does that otherwise this PR would have failed test I think.

bertinatto:
> > Once this is merged, CredentialsRequests will be created by CVO, so we'll need follow up PRs to CSI operator repositories to not create those objects themselves.
> 
> @bertinatto Can CCO handle cases where secret was already created? Looks like it already does that otherwise this PR would have failed test I think.

Yes, from what I've seen it can handle that.

manifests/03_credentials_request_aws.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: cloudcredential.openshift.io/v1
+kind: CredentialsRequest
```
gnufied, line 2: should this be called "ebs-credential-request"? In case we have to create separate credential request for EFS for example?

manifests/03_credentials_request_aws.yaml
```diff
@@ -0,0 +1,31 @@
...
+metadata:
+  name: openshift-cluster-csi-drivers
+  namespace: openshift-cloud-credential-operator
+spec:
+  secretRef:
+    name: aws-cloud-credentials
+    namespace: openshift-cluster-csi-drivers
```
gnufied, line 9: same applies to secret's name too.

bertinatto: We'll need to have a coordinated way of changing the secret name here and in the CSI driver operator repo.

If I change it here, the CI job will likely fail because the CSI driver is looking for this secret name.

manifests/03_credentials_request_aws.yaml
bertinatto: Changed to `aws-ebs-csi-driver-operator` to align with the other CCs which are using the operator name.

bertinatto:
/hold cancel

CSI driver operators should be changed after this PR merges.

manifests/03_credentials_request_aws.yaml
bertinatto: Renamed to `ebs-cloud-credentials` (similar to Manila and oVirt).

It shouldn't fail CI because for the time being the CSI driver operator is creating the secret itself.

PR #83, opened 2020-09-02
Bug 1874873: Move notUpgradeable check later to prevent z-stream upgrade failures

huffmanca:
Moves the v1alpha1 Snapshot CRD check to later in Reconcile, so that we can proceed with z-stream upgrades.


wking (approved):

bertinatto:
@huffmanca I noticed that the target branch is `release-4.5`; did we cherry-pick this from `master` (4.6)?

PR #85, opened 2020-09-04
Bug 1872549: Updated error message when SC is installed via CSI

huffmanca:
Currently if a platform has a CSI Driver but not a default SC (such as oVirt), we display the following error message:

> DefaultStorageClassControllerAvailable: No default StorageClass for this platform

This makes it a bit clearer that the SC is provided by a corresponding CSI driver. I've included the other supported platforms where we ship the CSI driver to future-proof the PR.

### DIFF
--- a/pkg/operator/defaultstorageclass/controller.go
+++ b/pkg/operator/defaultstorageclass/controller.go
@@ -29,6 +29,7 @@ const (
 )
 
 var unsupportedPlatformError = errors.New("unsupported platform")
+var supportedByCSIError = errors.New("only supported by a provided CSI Driver")
 
 // This Controller deploys a default StorageClass for in-tree volume plugins,
 // based on the underlying cloud (read from Infrastructure instance).
@@ -107,6 +108,18 @@ func (c *Controller) sync(ctx context.Context, syncCtx factory.SyncContext) erro
 				v1helpers.UpdateConditionFn(progressingCnd),
 			)
 			return updateErr
+		} else if syncErr == supportedByCSIError {
+			// Set Available=true, Progressing=false - everything is OK
+			// for this operator, but there may be work remaining for the
+			// external CSI Drivers.
+			availableCnd.Message = "StorageClass provided by supplied CSI Driver instead of the cluster-storage-operator"
+			availableCnd.Status = operatorapi.ConditionTrue
+
+			_, _, updateErr := v1helpers.UpdateStatus(c.operatorClient,
+				v1helpers.UpdateConditionFn(availableCnd),
+				v1helpers.UpdateConditionFn(progressingCnd),
+			)
+			return updateErr
 		}
 
 		// Set Available=false, Progressing=true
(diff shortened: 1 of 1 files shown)


pkg/operator/defaultstorageclass/controller.go
```diff
@@ -107,6 +109,18 @@ func (c *Controller) sync(ctx context.Context, syncCtx factory.SyncContext) erro
 				v1helpers.UpdateConditionFn(progressingCnd),
 			)
 			return updateErr
+		} else if syncErr == supportedByCSIError {
+			// Set Available=true, Progressing=false - everything is OK
+			// for this operator, but there may be work remaining for the
+			// external CSI Drivers.
+			availableCnd.Message = "StorageClass provided by supplied CSI Driver instead of the cluster-storage-operator"
+			availableCnd.Status = operatorapi.ConditionTrue
```
jsafrane, line 117: IMO it's not necessary to set any message - there is a default storage class somewhere and users IMO don't care where it comes from. Everything works as expected.

In other words, `syncStorageClass` could return nil (with a debug log message?)

pkg/operator/csidriveroperator/util.go
```diff
@@ -30,6 +31,8 @@ var (
 		"${NODE_DRIVER_REGISTRAR_IMAGE}", os.Getenv(envNodeDriverRegistrarImage),
 		"${LIVENESS_PROBE_IMAGE}", os.Getenv(envLivenessProbeImage),
 	)
+	// This should be updated with additional PlatformTypes as we ship new CSI Drivers
+	supportedCSIDrivers = []string{"OpenStack", "oVirt", "AWS"}
```
jsafrane, line 35: Use proper constants, such as `configv1.OpenStackPlatformType`

pkg/operator/defaultstorageclass/controller.go
```diff
@@ -177,6 +191,9 @@ func newStorageClassForCluster(infrastructure *configv1.Infrastructure) (*storag
 	case configv1.VSpherePlatformType:
 		return resourceread.ReadStorageClassV1OrDie(generated.MustAsset("storageclasses/vsphere.yaml")), nil
 	default:
+		if csi.SupportedCSIPlatform(infrastructure.Status.PlatformStatus.Type) {
```
jsafrane, line 194: Maybe this would be enough that's needed in this PR:
`case configv1.OVirtPlatformType: return nil`.

I don't like having some platforms listed in `supportedCSIDrivers` array and another list in this `switch`.

pkg/operator/defaultstorageclass/controller.go
huffmanca: The message was because QE indicated that it was slightly confusing reporting "No default StorageClass" while one was provided by a CSI driver.

pkg/operator/defaultstorageclass/controller.go
huffmanca: I removed the supportedCSIDrivers slice and just added the check for the OvirtPlatformType. If we remove the default StorageClass entries for other platforms we can simply update the call here.

PR #86, opened 2020-09-10
Bug 1866791: Bump library-go to get better operatorLogLevel validation

jsafrane:
* operatorLogLevel validation is done in https://github.com/openshift/library-go/pull/864.
* And get the latest client-go instead of -rc3 as side effect.

@openshift/storage


jsafrane:
/retest
unit test job now has go 1.15

PR #89, opened 2020-09-16
Bug 1879365: Move CSO namespace to lower runlevel

jsafrane:
cluster-storage-operator and cluster-csi-snapshot-controller-operator both run at runlevel 50 and in the same namespace. Create the namespace at runlevel 49, so the namespace is created only once and the authoritative version is in this repo.

The namespace creation will be removed from cluster-csi-snapshot-controller-operator in a subsequent PR.

### DIFF
--- a/manifests/0000_49_cluster-storage-operator_01_operator_namespace.yaml
+++ b/manifests/0000_49_cluster-storage-operator_01_operator_namespace.yaml
@@ -1,0 +1,9 @@
+# Create the namespace at runlevel 49, so cluster-csi-snapshot-controller-operator (running at level 50) can use it.
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    openshift.io/node-selector: ""
+  labels:
+    openshift.io/cluster-monitoring: "true"
+  name: openshift-cluster-storage-operator
--- a/manifests/01_operator_namespace.yaml
+++ b/manifests/01_operator_namespace.yaml
@@ -1,6 +1,0 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-  annotations:
-    openshift.io/node-selector: ""
-  name: openshift-cluster-storage-operator


jsafrane:
cc @openshift/storage

manifests/0000_50_cluster-storage-operator_01_operator_namespace.yaml
```diff
@@ -1,3 +1,4 @@
+# Create the namespace at runlevel 49, so cluster-csi-snapshot-controller-operator (running at level 50) can use it.
 apiVersion: v1
 kind: Namespace
 metadata:
```
wking, line 4: Was [the `openshift.io/cluster-monitoring: "true"` label][1] important?

[1]: https://github.com/openshift/cluster-csi-snapshot-controller-operator/pull/49/files#diff-c90f942468bb6270eaab50e76232634eL8

manifests/0000_50_cluster-storage-operator_01_operator_namespace.yaml
```diff
@@ -1,3 +1,4 @@
+# Create the namespace at runlevel 49, so cluster-csi-snapshot-controller-operator (running at level 50) can use it.
```
bertinatto, line 1: To create the NS at runlevel 49, shouldn't you rename it to `manifests/0000_49_cluster-storage-operator_01_operator_namespace.yaml` instead?

manifests/0000_50_cluster-storage-operator_01_operator_namespace.yaml
jsafrane: Restored

manifests/0000_50_cluster-storage-operator_01_operator_namespace.yaml
jsafrane: Fixed

bertinatto (commented):
/lgtm

PR #91, opened 2020-09-21
Bug 1882394: Delete lock created by CSO 4.6

bertinatto:
Leader-for-life leader election causes problems when downgrading **from** CSO 4.6, which uses library-go and doesn't set an `ownerReference` in the `ConfigMap` metadata.

This patch identifies if the lock was created by CSO 4.6 and, if so, deletes it so that CSO 4.5 can become the leader using the leader-for-life model.

More on the the problem:

On a downgrade from 4.5 to 4.6 scenario, the leader election `ConfigMap` is created by CSO 4.6 (library-go) and it doesn't have any `ownerReference` set.

However, CSO 4.5 uses `operator-sdk` with leader-for-life election approach, and it expects that the `ConfigMap` either doesn't exist or exists but has an `ownerReference` set (in order to know whether the lock belongs to it or not).

Since it doesn't find any, it tries to create one, but it gets back an error stating the the object already exists. Therefore, CSO 4.5 thinks that the lock belongs to somebody else and it never starts. This goes on forever and CSO 4.5 never become the leader.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -6,13 +6,20 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-storage-operator/pkg/apis"
 	"github.com/openshift/cluster-storage-operator/pkg/controller"
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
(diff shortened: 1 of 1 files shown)


bertinatto:
/hold
To be tested.

cmd/manager/main.go
```diff
@@ -47,8 +96,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Delete incompatible ConfigMap lock if it exists
+	lockName := "cluster-storage-operator-lock"
+	err = tryDeleteIncompatibleLock(cfg, lockName)
+	if err != nil {
+		// We don't exit if an error happens, but we let the operator
+		// try to become the leader below. If it doesn't succeed, then we exit
```
jsafrane, line 104: In https://bugzilla.redhat.com/show_bug.cgi?id=1877316 we saw 4.5 CSO trying to become leader for 3 hours without exiting. There should be either timeout for `leader.Become` or CSO should exit here.

bertinatto: I was expecting that on line 109 below it would `os.Exit(1)` if CSO couldn't become the leader, but operator-sdk doesn't return an error the ConfigMap already exists:

https://github.com/bertinatto/cluster-storage-operator/blob/1b1fae52bda9667b7b9bf7da7397c88a82faa4cc/vendor/github.com/operator-framework/operator-sdk/pkg/leader/leader.go#L124-L131

bertinatto: Fixed, PTAL

bertinatto:
/retest
`[sig-network]` flakes

cmd/manager/main.go
```diff
@@ -47,8 +98,22 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Delete incompatible ConfigMap lock if it exists
+	lockName := "cluster-storage-operator-lock"
+	errWait := wait.PollImmediate(time.Second*10, time.Minute*5, func() (bool, error) {
+		err := tryDeleteIncompatibleLock(cfg, lockName)
+		if err != nil {
+			return false, err
```
jsafrane, line 106: Please log the error. 

* `log.Error(errWait, ...` below logs `Deadline exceeded` or some timeout message, not the API error
* It would be better to get some error logged before 5 minute timeout, users are not that patient.

bertinatto: Added log

jsafrane:
Using my team lead superpowers to add valid-bug label: there is nothing to fix in 4.6

bertinatto:
/retest
resize flake

bertinatto:
/retest
```
[sig-apps][Feature:DeploymentConfig] deploymentconfigs adoption will orphan all RCs and adopt them back when recreated 

[sig-storage] In-tree Volumes [Driver: aws] [Testpattern: Inline-volume (ext3)] volumes should store data
```

jsafrane:
/hold cancel
Most of the test pass and CSO is installed.

PR #93, opened 2020-09-29
Bug 1878040: Copy CRD manifest from openshift/api

jsafrane:
openshift/api has authoritative version of CSO CRD, so copy it from there hen building an image. Bump of openshift/api is enough to get a new version of the CRD.

Re-using bug 1878040, it's slightly related, see https://github.com/openshift/cluster-storage-operator/pull/92

@openshift/storage


bertinatto (commented):
/lgtm

PR #94, opened 2020-10-05
Bump OpenShift API to include Cinder CSI driver

Fedosin:


bertinatto (commented):
This LGTM.

@jsafrane just to make sure I'm not missing anything: could you take a look as well?

PR #97, opened 2020-10-23
add current profile annotations to CVO manifests

guillaumerose:
This is matches openshift/enhancements#510 and doesn't change existing behavior


deads2k:
This is consistent with our default opt-in profile

/lgtm

/assign @jsafrane 

for approval.

PR #100, opened 2020-11-04
Bug 1894025: Add annotations to cluster-csi-drivers namespace

bertinatto:
CC @openshift/storage

### DIFF
--- a/assets/csidriveroperators/aws-ebs/01_namespace.yaml
+++ b/assets/csidriveroperators/aws-ebs/01_namespace.yaml
@@ -2,3 +2,6 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: openshift-cluster-csi-drivers
+  annotations:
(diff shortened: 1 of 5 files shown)


assets/csidriveroperators/aws-ebs/01_namespace.yaml
```diff
@@ -2,3 +2,6 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: openshift-cluster-csi-drivers
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
```
bertinatto, line 6: This line is to match https://github.com/openshift/cluster-storage-operator/commit/d558dbc19a79aa2f3cc758f2d49a06faabf3d4c2

bertinatto: In theory we could simply remove these files from `assets` because the `openshift-cluster-csi-drivers` namespace is created by CVO.

PR #101, opened 2020-11-16
Add GCP-PD CSI Operator to CSO

huffmanca:
Adds the GCP-PD CSI Operator to CSO.


assets/csidriveroperators/gcp-pd/07_deployment.yaml
```diff
@@ -0,0 +1,51 @@
...
+    metadata:
+      labels:
+        name: gcp-pd-csi-driver-operator
+    spec:
+      containers:
+      - args:
+        - start
+        env:
+        - name: DRIVER_IMAGE
+          value: quay.io/openshift/origin-gcp-pd-csi-driver:latest
```
bertinatto, line 22: Is there any reason for not using the placeholders (e.g., `${DRIVER_IMAGE}`) now?

pkg/operator/csidriveroperator/csioperatorclient/gcp-pd.go
```diff
@@ -0,0 +1,39 @@
...
+const (
+	GCPPDCSIDriverName          = "pd.csi.storage.gke.io"
+	envGCPPDDriverOperatorImage = "GCP_PD_DRIVER_OPERATOR_IMAGE"
+	envGCPPDDriverImage         = "GCP_PD_DRIVER_IMAGE"
+)
+
+func GetGCPPDCSIOperatorConfig() CSIOperatorConfig {
+	pairs := []string{
+		"${OPERATOR_IMAGE}", os.Getenv(envGCPPDDriverOperatorImage),
+		"${DRIVER_IMAGE}", os.Getenv(envGCPPDDriverImage),
```
bertinatto, line 19: These env vars aren't set anywhere else in this PR, are they?

You should specify them in `manifests/09_deployment.yaml`

assets/csidriveroperators/gcp-pd/01_namespace.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-cluster-csi-drivers
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    openshift.io/node-selector: ""
```
bertinatto, line 7: The `verify` job is failing because the bindata isn't updated, so we need a `make update` to fix that.

assets/csidriveroperators/gcp-pd/07_deployment.yaml
huffmanca: There isn't. This has been updated.

pkg/operator/csidriveroperator/csioperatorclient/gcp-pd.go
huffmanca: Looking at the other operators I don't see any of these doing a similar method? They're all specified in the corresponding *.go files?

bertinatto: I'm not sure I understand your question...

At lines 18 and 19 the code is reading `GCP_PD_DRIVER_OPERATOR_IMAGE` and `GCP_PD_DRIVER_IMAGE` env vars. However, they are not set anywhere, so they are going to be empty.

We should set them in the CSO deployment, like we do for AWS for example: https://github.com/openshift/cluster-storage-operator/blob/master/manifests/09_deployment.yaml#L50-L53

huffmanca: I misunderstood what you were saying. I'll add these variables once they're available on quay.

From speaking with Thiago this should happen soon.

manifests/09_deployment.yaml
```diff
@@ -51,6 +51,10 @@ spec:
             value: quay.io/openshift/origin-aws-ebs-csi-driver-operator:latest
           - name: AWS_EBS_DRIVER_IMAGE
             value: quay.io/openshift/origin-aws-ebs-csi-driver:latest
+          - name: GCP_PD_DRIVER_OPERATOR_IMAGE
+            value: quay.io/huffmanca/gcp-operator:latest
+          - name: GCP_PD_DRIVER_IMAGE
+            value: quay.io/huffmanca/gcp:latest
```
bertinatto, line 57: You can use `quay.io/openshift` here. This will be replaced by the cluster's own image.

huffmanca:
Testing this with personal images allows volumes to be dynamically provisioned from the created CSI StorageClass, and all pods are running successfully. Removing WIP.

jsafrane:
/lgtm
GCE installation passed

PR #102, opened 2020-11-18
Add IBM Cloud managed annotations to CVO manifests

csrwng:
This matches https://github.com/openshift/enhancements/pull/445 and does not change behavior.

It requires a bump of openshift/api so we can include changes in https://github.com/openshift/api/pull/781


jsafrane:
@csrwng can you please rebase?
/approve

csrwng:
Thanks, rebased

PR #103, opened 2020-11-25
Add OpenStack Cinder driver startup

Fedosin:


jsafrane:
/lgtm
I can see Cinder CSI driver + the operator is running in e2es

PR #104, opened 2020-11-27
Bug 1902552: Fix cinder-csi images

mandre:
The images were hardcoded to quay.io registries. They should instead be
templated to use values from the release payload.

Bindata generated with `make update-bindata`.


Fedosin (approved):
>_< 
Sorry, my bad... These images are not used for real, we replace them when we build the release image. But yeah, there should not be any personal images, only those from openshift namespace. 

/lgtm

mandre:
The platform running the openstack tests seems to be having a rough time.
/retest

jsafrane:
/lgtm
I checked the logs, Cluster-storage-operator is installed together with Cinder CSI driver

PR #109, opened 2020-12-03
support custom CA bundle for AWS API

staebler:
When a cluster is installed in a AWS C2S region, access to the AWS API requires a custom CA bundle for trust. The custom CA bundle is read from the "ca-bundle.pem" key of the kube-cloud-config ConfigMap in the openshift-config-managed namespace.

To make the custom CA bundle available for use by the EBS CSI driver, the operator will sync the kube-cloud-config ConfigMap into the openshift-cluster-csi-drivers namespace. The EBS CSI driver operator is responsible for adjusting the deployment for the driver to mount the CA bundle and add the appropriate environment variable.

To inform the EBS CSI driver operator that there is a custom CA bundle, the operator will set the CA_BUNDLE_CONFIG_MAP environment variable to the name of the kube-cloud-config ConfigMap.

https://issues.redhat.com/browse/CORS-1584


staebler:
Moving the ConfigMap syncing to the AWS EBS operator instead.

/close

PR #106, opened 2020-12-03
Bug 1915473: Annotate manifests for single-node-developer cluster profile

cfergeau:
This partially implements phase 1 of https://github.com/openshift/enhancements#482
and does not change behavior. Initially, all cluster-storage-operator
manifests are included in the single-node-developer cluster profile.
Follow-on PRs may exclude any of these that are not needed in the
profile.

PR #113, opened 2020-12-06
add 'single-node-production-edge' annotations to CVO manifests.

osherdp:
This adds annotations for the single-node-production-edge cluster profile. There's a growing requirement from several customers to enable creation of single-node (not high-available) Openshift clusters.
In stage one (following openshift/enhancements#504) there should be no implication on components logic.
In the next stage, the component's behavior will match a non high-availability profile if the customer is specifically interested in one.
This PR is separate from the 'single-node-developer' work, which will implement a different behavior and is currently on another stage of implementation.

For more info, please refer to the enhancement link and participate in the discussion.

### DIFF
--- a/manifests/0000_49_cluster-storage-operator_01_operator_namespace.yaml
+++ b/manifests/0000_49_cluster-storage-operator_01_operator_namespace.yaml
@@ -6,6 +6,7 @@ metadata:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     openshift.io/node-selector: ""
+    include.release.openshift.io/single-node-production-edge: "true"
   labels:
     openshift.io/cluster-monitoring: "true"
   name: openshift-cluster-storage-operator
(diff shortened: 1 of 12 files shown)


jsafrane:
@osherdp, do you expect a single node cluster to run in a cloud? On bare metal, cluster-storage-operator does not do anything useful and could be omitted.

jsafrane:
And please sync with @cfergeau and #106

osherdp:
@jsafrane indeed, not relevant on bare-metal
for CI purposes we do need to run SNO on cloud

bertinatto (commented):
/lgtm

jsafrane:
> for CI purposes we do need to run SNO on cloud

You can run it on a cloud and pretend it's a bare metal cluster, i.e. not to use any cloud-y feature such as persistent storage. Why would you test cloud persistent volumes if the end product won't support them?

osherdp:
Shifting strategy based on https://github.com/openshift/enhancements/pull/560

PR #118, opened 2021-01-05
Bug 1907812: Use separate RBAC objects for AWS CA bundle retrieval

jsafrane:
When downgrading 4.7 cluster to 4.6, we must make sure that the 4.7 operator still can read AWS configuration (`openshift-config-managed/kube-cloud-config` ConfigMap) when CVO / OCS already downgraded the RBAC rules to 4.6 version. Otherwise the operator gets degraded.

Therefore use a separate RBAC objects, they won't be removed when downgrading to 4.6 at all.

@openshift/storage


assets/csidriveroperators/aws-ebs/07_clusterrole_aws_config.yaml
```diff
@@ -0,0 +1,15 @@
+# Allow reading openshift-config-managed/kube-cloud-config ConfigMap
+# with CA bundle.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
bertinatto, line 4: Since the goal here is to allow reading a specific ConfigMap in a specific namespace, could we convert this `ClusterRole` into a `Role` and specify the namespace and the resource name? Like:
```
metadata:
  namespace: openshift-config-managed
(...)
  resources: 
  - configmaps
  resourceNames:
  - kube-cloud-config
  (...)
```

jsafrane: I tried `resourceNames:`, but CSO failed to `watch` - it watches all configmaps in the namespace. I'll try `Role` referring to a ServiceAccount in a different namespace.

jsafrane: Ok, it seems it does something - at least AWS EBS operator is not complaining about RBACs.

bertinatto (commented):
/lgtm

PR #119, opened 2021-01-06
Bug 1909978: specify availability zone for OpenStack SC

Fedosin:
The absence of "availability" parameter in the storage class prevents users from setting "ignore-volume-az = yes" in the cloud provider config.

### DIFF
--- a/assets/storageclasses/openstack.yaml
+++ b/assets/storageclasses/openstack.yaml
@@ -5,6 +5,8 @@ metadata:
   annotations:
     storageclass.kubernetes.io/is-default-class: "true"
 provisioner: kubernetes.io/cinder
+parameters:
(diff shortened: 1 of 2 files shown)


assets/storageclasses/openstack.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
   annotations:
     storageclass.kubernetes.io/is-default-class: "true"
 provisioner: kubernetes.io/cinder
+parameters:
+  availability: nova
```
mandre, line 9: What happens when the nova availability zone does not exist?

Fedosin:
We can't updated existing storage classes, because it prevents upgrades. Therefore I close this PR

PR #126, opened 2021-01-14
Bug 1904503: Add prometheus alerts for vsphere

gnufied:
Depends on https://github.com/openshift/vsphere-problem-detector/pull/24/files

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1904503

### DIFF
--- a/assets/vsphere_problem_detector/12_prometheusrules.yaml
+++ b/assets/vsphere_problem_detector/12_prometheusrules.yaml
@@ -15,11 +15,11 @@ spec:
         labels:
           severity: warning
         annotations:
-          message: "Vsphere node health checks are failing on {{ $labels.node }} with {{ $labels.check }}"
+          message: "VSphere health check {{ $labels.check }} is failing on {{ $labels.node }}."
       - alert: VSphereOpenshiftClusterHealthFail
         expr: vsphere_cluster_check_errors == 1
         for: 10m
         labels:
-          severity: critical
+          severity: warning
         annotations:
-          message: "VSpehre cluster health checks are failing with {{ $labels.check }}"
+          message: "VSphere cluster health checks are failing with {{ $labels.check }}"
--- a/assets/vsphere_problem_detector/12_prometheusrules.yaml
+++ b/assets/vsphere_problem_detector/12_prometheusrules.yaml
@@ -1,0 +1,25 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: vsphere-problem-detector
+  namespace: openshift-cluster-storage-operator
+  labels:
+    role: alert-rules
+spec:
+  groups:
+    - name: vsphere-problem-detector.rules
+      rules:
+      - alert: VSphereOpenshiftNodeHealthFail
+        expr:  vsphere_node_check_errors == 1
+        for: 10m
+        labels:
+          severity: warning
+        annotations:
+          message: "Vsphere node health checks are failing on {{ $labels.node }} with {{ $labels.check }}"
+      - alert: VSphereOpenshiftClusterHealthFail
+        expr: vsphere_cluster_check_errors == 1
+        for: 10m
+        labels:
+          severity: critical
+        annotations:
+          message: "VSpehre cluster health checks are failing with {{ $labels.check }}"
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/openshift/client-go v0.0.0-20200827190008-3062137373b5
 	github.com/openshift/library-go v0.0.0-20200909144351-f29d76719396
 	github.com/prometheus-operator/prometheus-operator v0.44.1
+	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.44.1
 	github.com/prometheus/client_golang v1.8.0
 	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
--- a/pkg/generated/bindata.go
+++ b/pkg/generated/bindata.go
@@ -3453,14 +3453,14 @@ spec:
         labels:
           severity: warning
         annotations:
-          message: "Vsphere node health checks are failing on {{ $labels.node }} with {{ $labels.check }}"
+          message: "VSphere health check {{ $labels.check }} is failing on {{ $labels.node }}."
       - alert: VSphereOpenshiftClusterHealthFail
         expr: vsphere_cluster_check_errors == 1
         for: 10m
         labels:
-          severity: critical
+          severity: warning
         annotations:
-          message: "VSpehre cluster health checks are failing with {{ $labels.check }}"
+          message: "VSphere cluster health checks are failing with {{ $labels.check }}"
 `)
 
 func vsphere_problem_detector12_prometheusrulesYamlBytes() ([]byte, error) {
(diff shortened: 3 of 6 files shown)


assets/vsphere_problem_detector/12_prometheusrules.yaml
```diff
@@ -0,0 +1,25 @@
...
+          severity: warning
+        annotations:
+          message: "Vsphere node health checks are failing on {{ $labels.node }} with {{ $labels.check }}"
+      - alert: VSphereOpenshiftClusterHealthFail
+        expr: vsphere_cluster_check_errors == 1
+        for: 10m
+        labels:
+          severity: critical
+        annotations:
+          message: "VSpehre cluster health checks are failing with {{ $labels.check }}"
```
jsafrane, line 25: typo: VSpehre

assets/vsphere_problem_detector/12_prometheusrules.yaml
```diff
@@ -0,0 +1,25 @@
...
+  groups:
+    - name: vsphere-problem-detector.rules
+      rules:
+      - alert: VSphereOpenshiftNodeHealthFail
+        expr:  vsphere_node_check_errors == 1
+        for: 10m
+        labels:
+          severity: warning
+        annotations:
+          message: "Vsphere node health checks are failing on {{ $labels.node }} with {{ $labels.check }}"
```
jsafrane, line 18: Is it OK to alert on each node separately? This may be quite spammy if all nodes are equally bad.

gnufied: @openshift/openshift-team-monitoring

assets/vsphere_problem_detector/12_prometheusrules.yaml
gnufied: fixed

assets/vsphere_problem_detector/12_prometheusrules.yaml
```diff
@@ -0,0 +1,25 @@
...
+        for: 10m
+        labels:
+          severity: warning
+        annotations:
+          message: "VSphere node health checks are failing on {{ $labels.node }} with {{ $labels.check }}"
+      - alert: VSphereOpenshiftClusterHealthFail
+        expr: vsphere_cluster_check_errors == 1
+        for: 10m
+        labels:
+          severity: critical
```
jsafrane, line 23: The other one is `warning`, do we really want it critical? I'd start as low as possible, we don't know how many clusters are going to report the alert after upgrade to 4.7

assets/vsphere_problem_detector/12_prometheusrules.yaml
```diff
@@ -0,0 +1,25 @@
...
+        annotations:
+          message: "VSpehre cluster health checks are failing with {{ $labels.check }}"
```
jsafrane, line 25: still typo: VSpehre

assets/vsphere_problem_detector/12_prometheusrules.yaml
```diff
@@ -0,0 +1,25 @@
...
+  groups:
+    - name: vsphere-problem-detector.rules
+      rules:
+      - alert: VSphereOpenshiftNodeHealthFail
+        expr:  vsphere_node_check_errors == 1
+        for: 10m
+        labels:
+          severity: warning
+        annotations:
+          message: "VSphere node health checks are failing on {{ $labels.node }} with {{ $labels.check }}"
```
jsafrane, line 18: This sounds better to me: "VSphere health check {{ $labels.check }} is failing on node {{ $labels.node }}"

But who I am to comment on other's English style :-)

gnufied: You win this round. I renamed. :-)

assets/vsphere_problem_detector/12_prometheusrules.yaml
gnufied: err. I fixed wrong typo. fix this one. sorry.

assets/vsphere_problem_detector/12_prometheusrules.yaml
gnufied: changed this to warning as well.

assets/vsphere_problem_detector/12_prometheusrules.yaml
gnufied: According to monitoring team:

> e.g you have 2 nodes for which vsphere_node_check_errors = 1, you'll receive 1 notification containing 2 alerts firing along with the message for each failing node 

I think it might be okay. It is not unusual for a 100 node cluster to all going wrong at once, but I think the entire class can be disabled at once and hence should be okay. If we remove node name from here, then it will be less useful I think because we can't tell from which node these alerts are coming.

assets/vsphere_problem_detector/12_prometheusrules.yaml
```diff
@@ -0,0 +1,25 @@
...
+  name: vsphere-problem-detector
+  namespace: openshift-cluster-storage-operator
+  labels:
+    role: alert-rules
+spec:
+  groups:
+    - name: vsphere-problem-detector.rules
+      rules:
+      - alert: VSphereOpenshiftNodeHealthFail
+        expr:  vsphere_node_check_errors == 1
```
lilic, line 13: What type is this metric? A counter or gauge? Quick search could not find it in the operator.

lilic (commented):
Thank you for ping! 🎉 

Just curious what type is, bit worried this will falsely be firing always depending on the type of the metric.

assets/vsphere_problem_detector/12_prometheusrules.yaml
jsafrane: It's gauge now, https://github.com/openshift/vsphere-problem-detector/pull/24
(used to be counter yesterday, hard to alert on it).

jsafrane:
lgtm-ish, waiting for @lilic's approval / additional comments.

assets/vsphere_problem_detector/12_prometheusrules.yaml
simonpasquier: One small comment to improve the alerting rule: in the current version, a failed scrape by Prometheus would resolve the alert if it was firing previously.
To protect against it, you can use `min_over_time()` like this

```suggestion
        expr:  min_over_time(vsphere_node_check_errors[5m]) == 1
```

see https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0

lilic: > (used to be counter yesterday, hard to alert on it).

Nice thanks! 

Agreed with Simons suggestion, otherwise looks good to me.

gnufied: why `min_over_time` - shouldn't this be `max_over_time`?  I would think if a scrape failed and a value is missing on time t1 then it would replace with some kind of sentinel value (`0`?  - I don't know prometheus very well and how it fills the holes in the data) .

simonpasquier: If the target is down (up == 0) then Prometheus will mark the `vsphere_node_check_errors` metric as stale (meaning it doesn't exist anymore). On the next evaluation of the alerting rule, the result from the rule's expression would be "no data" so Prometheus will consider that the alert is resolved.

PR #121, opened 2021-01-11
Bug 1912945: Add RBAC rules to read proxy resource

bertinatto:
This will be necessary so that CSI driver operators can read the proxy object.

CC @openshift/storage

PR #127, opened 2021-01-18
Bug 1912945: Add RBAC to allow operator to update CR's spec

bertinatto:
The Config Observer controller from CSI driver operators `1)` watches the `proxies` resource and `2)` patches/updates the `CR.Spec.ObservedConfig`.

PR https://github.com/openshift/cluster-storage-operator/pull/121 introduced the RBAC rule for `#1`, but it missed the rule for `#2`.

CC @openshift/storage


tsmetana (approved):
/lgtm

PR #130, opened 2021-01-19
Bug 1905141: Fix alerting rule of vsphere-problem-detector

jsafrane:
When vsphere-problem-detector's prometheus endpoint is down from any reason (e.g. update), Prometheus will report "no data" for `vsphere_*_check_errors` and therefore clear any current alert.

Using `min_over_time[5m]` will ensure that such "no data" holes are skipped.


jsafrane:
/close 
dup of https://github.com/openshift/cluster-storage-operator/pull/128

PR #131, opened 2021-01-19
Bug 1910581: CSO shouldn't overwrite clustercsidriver objects

bertinatto:
With this PR CSO will no longer update/patch the CR.

If the CR doesn't exist, CSO will create it. However, if the already CR exists, CSO won't update it with the config that it sees fit.

This is a requirement for the CSI Config Observer Controller used in CSI driver operators. That controller watches `proxies` resources and writes the observed config in `CR.Spec.observedConfig`. Without this PR, CSO would overwrite that field.

CC @openshift/storage

### DIFF
--- a/pkg/operator/csidriveroperator/crcontroller.go
+++ b/pkg/operator/csidriveroperator/crcontroller.go
@@ -14,14 +14,12 @@ import (
 	"github.com/openshift/cluster-storage-operator/pkg/operator/csidriveroperator/csioperatorclient"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
-	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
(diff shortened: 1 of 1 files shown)


bertinatto:
/wip
Still checking if this patch fixes the issue.

pkg/operator/csidriveroperator/crcontroller.go
```diff
@@ -200,25 +178,7 @@ func (c *CSIDriverOperatorCRController) applyClusterCSIDriver(requiredOriginal *
 		return nil, false, err
 	}
 
-	modified := resourcemerge.BoolPtr(false)
-	existingCopy := existing.DeepCopy()
-
-	resourcemerge.EnsureObjectMeta(modified, &existingCopy.ObjectMeta, required.ObjectMeta)
```
gnufied, line 206: I think we might want to merge metadata of an object in-case there are some cluster-wide annotations applied to these CRs, but if we come to that we can also push and update and do metadata merging later on (so I am bit torn).

PR #133, opened 2021-01-29
Bug 1883538: Only include the relevant drivers in relatedObjects

huffmanca:
Adjusts the relatedObjects of the `storage` CO to only include the relevant CSI Drivers instead of every driver.

### DIFF
--- a/pkg/operator/csidriveroperator/driverstarter.go
+++ b/pkg/operator/csidriveroperator/driverstarter.go
@@ -24,6 +24,10 @@ const (
 	infraConfigName = "cluster"
 )
 
+var (
+	relatedObjects []configv1.ObjectReference
+)
+
 // This CSIDriverStarterController starts CSI driver controllers based on the
 // underlying cloud and removes it from OLM. It does not install anything by
 // itself, only monitors Infrastructure instance and starts individual
@@ -35,8 +39,7 @@ type CSIDriverStarterController struct {
 	versionGetter  status.VersionGetter
 	targetVersion  string
 	eventRecorder  events.Recorder
-
-	controllers []csiDriverControllerManager
+	controllers    []csiDriverControllerManager
 }
 
 type csiDriverControllerManager struct {
@@ -61,6 +64,7 @@ func NewCSIDriverStarterController(
 		targetVersion:  targetVersion,
 		eventRecorder:  eventRecorder.WithComponentSuffix("CSIDriverStarter"),
 	}
+	relatedObjects = []configv1.ObjectReference{}
 
 	// Populating all CSI driver operator ControllerManagers here simplifies
 	// the startup a lot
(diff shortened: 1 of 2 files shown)


pkg/operator/starter.go
```diff
@@ -82,7 +74,8 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 		versionGetter,
 		status.VersionForOperandFromEnv(),
 		controllerConfig.EventRecorder,
-		csiDriverConfigs)
+		csiDriverConfigs,
+		clusterOperatorStatus)
```
gnufied, line 78: I think it might be cleaner to write this code like:

```
	csiDriverController := csidriveroperator.NewCSIDriverStarterController(
		clients,
		resync,
		versionGetter,
		status.VersionForOperandFromEnv(),
		controllerConfig.EventRecorder,
		csiDriverConfigs)
	clusterOperatorStatus.WithRelatedObjectsFunc(csiDriverController.RelatedObjectFunc)
```

and in CSI driver starter:

```
func (c *CSIDriverStarterController) RelatedObjectFunc() func() (isset bool, objs []configv1.ObjectReference) {
	return func() (isset bool, objs []configv1.ObjectReference) {
		if len(c.relatedObjects) == 0 {
			return false, c.relatedObjects
		}
		return true, c.relatedObjects
	}
}
```

huffmanca: I've updated the PR to take a similar approach. From initial testing, this works on AWS. I want to confirm other platforms (and test longer to ensure we're not flapping the relatedObjects).

huffmanca: Just FYI - I've tested thi son a GCP cluster, and it includes the following relatedObject entry:

```
    {
      "group": "operator.openshift.io",
      "name": "pd.csi.storage.gke.io",
      "resource": "clustercsidrivers"
    },
```

pkg/operator/csidriveroperator/driverstarter.go
```diff
@@ -112,6 +116,18 @@ func (c *CSIDriverStarterController) sync(ctx context.Context, syncCtx factory.S
...
+				Resource: "clustercsidrivers",
+				Name:     ctrl.operatorConfig.CSIDriverName,
+			})
+			// We need a special check for Manila to add the OLM namespace
+			if platform == configv1.OpenStackPlatformType {
+				relatedObjects = append(relatedObjects, configv1.ObjectReference{
+					Resource: "namespaces",
+					Name:     "openshift-manila-csi-driver",
+				})
+			}
```
gnufied, line 130: Do we still need to carry some code for manila OLM thingy? I think all such operators would have already been migrated. cc @jsafrane

huffmanca: This has been removed.

PR #138, opened 2021-02-18
Propagate log level to GCE and Cinder CSI driver operators

jsafrane:
Looks like oversight, GCE PD and Cinder CSI driver operators did not get logLevel parameter.

@bertinatto, PTAL

### DIFF
--- a/assets/csidriveroperators/openstack-cinder/07_deployment.yaml
+++ b/assets/csidriveroperators/openstack-cinder/07_deployment.yaml
@@ -17,6 +17,7 @@ spec:
       containers:
       - args:
         - start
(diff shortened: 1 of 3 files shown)


assets/csidriveroperators/openstack-cinder/07_deployment.yaml
```diff
@@ -17,6 +17,7 @@ spec:
       containers:
       - args:
         - start
+        - -v=${LOG_LEVEL}
```
bertinatto, line 20: We need to update the assets

bertinatto (commented):
/lgtm

PR #140, opened 2021-03-10
Include azure disk csi driver

huffmanca:
Includes the Azure Disk CSI Driver to be deployed via CSO. This is dependent on the images being built by ART and available at release.

I've tested this on an Azure cluster, and I'm able to dynamically provision volumes and write to them using the deployed CSI driver without issue.


huffmanca:
I'll go ahead and close this PR, as Jan is doing this in https://github.com/openshift/cluster-storage-operator/pull/148 .

PR #145, opened 2021-03-30
Prepare for metrics collection in CSI driver operators

jsafrane:
This is preparation for metrics collection in the CSI driver operators themselves.

* The CSI driver namespace must have label `openshift.io/cluster-monitoring` to be observed by Prometheus
* All CSI driver operators and their kube-rbac-proxies should be able to create `TokenReview` to authenticated Prometheus when it's collecting the metrics.
* When at it, I removed namespace creation from CSI driver operators. `openshift-cluster-csi-drivers` is created by CVO, to have it available when CCO creates cloud credentials there.


assets/csidriveroperators/aws-ebs/05_clusterrole.yaml
```diff
@@ -260,3 +260,10 @@ rules:
   - get
   - list
   - watch
+# Allow kube-rbac-proxy to create TokenReview to be able to authenticate Prometheus when collecting metrics
+- apiGroups:
+  - "authentication.k8s.io"
+  resources:.
+  - "tokenreviews"
+  verbs:
+  - "create"
```
gnufied, line 269: IIRC we also needed `subjectaccessreviews` for similar purpose. Is that still true or am I missing something?

assets/csidriveroperators/aws-ebs/05_clusterrole.yaml
```diff
@@ -260,3 +260,10 @@ rules:
   - get
   - list
   - watch
+# Allow kube-rbac-proxy to create TokenReview to be able to authenticate Prometheus when collecting metrics
+- apiGroups:
+  - "authentication.k8s.io"
+  resources:.
```
gnufied, line 266: typo

assets/csidriveroperators/manila/05_clusterrole.yaml
```diff
@@ -250,3 +250,10 @@ rules:
   - 'clustercsidrivers/status'
   verbs:
   - '*'
+# Allow kube-rbac-proxy to create TokenReview to be able to authenticate Prometheus when collecting metrics
+- apiGroups:
+  - "authentication.k8s.io"
+  resources:.
```
gnufied, line 256: same

pkg/generated/bindata.go
```diff
@@ -2733,6 +2771,13 @@ rules:
   # The Config Observer controller updates the CR's spec
   - update
   - patch
+# Allow kube-rbac-proxy to create TokenReview to be able to authenticate Prometheus when collecting metrics
+- apiGroups:
+  - "authentication.k8s.io"
+  resources:.
```
gnufied, line 2777: same as other places too.

assets/csidriveroperators/aws-ebs/05_clusterrole.yaml
gnufied: alright - lets see if that works. May be something special about going through a proxy.

assets/csidriveroperators/aws-ebs/05_clusterrole.yaml
jsafrane: Fixed

assets/csidriveroperators/aws-ebs/01_namespace.yaml
```diff
@@ -5,3 +5,5 @@ metadata:
   annotations:
     include.release.openshift.io/self-managed-high-availability: "true"
     openshift.io/node-selector: ""
+  labels:
+    openshift.io/cluster-monitoring: "true"
```
bertinatto, line 9: We should add the label [in this file](https://github.com/openshift/cluster-storage-operator/blob/master/manifests/02_csi_driver_operators_namespace.yaml), since it's CVO who creates this namespace.

In theory we could remove these asset files from here.

jsafrane: Hold on, why do we create the namespace unconditionally? It has no use on bare metal.

jsafrane: from git log:
```
Let CVO create namespace for CSI driver operators.
   
So that the openshift-cluster-csi-drivers namespace exists
when the cloud-credential-operator deploys the secret.
```
Good catch! I removed the namespace creation from all CSI driver operators.

assets/csidriveroperators/aws-ebs/01_namespace.yaml
```diff
@@ -1,7 +0,0 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: openshift-cluster-csi-drivers
-  annotations:
-    include.release.openshift.io/self-managed-high-availability: "true"
-    openshift.io/node-selector: ""
```
bertinatto, line 7: Next time we are here we should rename the other files (the first asset file is prefixed with `02_`.

bertinatto (commented):
/lgtm

PR #149, opened 2021-04-06
Remove "vmware" prefix from CSI operator and driver

bertinatto:
This is the third step described here: https://docs.ci.openshift.org/docs/how-tos/onboarding-a-new-component/#removing-or-renaming-an-image-from-an-openshift-release-image

CC @openshift/storage


bertinatto:
/hold
/retitle Remove "vmware" prefix from CSI operator and driver

jsafrane:
I can see the CSI driver is installed in vsphere-csi job, so the PR did not break it.

PR #151, opened 2021-04-08
Pass KUBE_RBAC_PROXY_IMAGE to CSI driver operators

jsafrane:
So they can use the right kube-rbac-proxy image in their Pods.

cc @openshift/storage


bertinatto (commented):
/lgtm

PR #152, opened 2021-04-08
Degrade the cluster when an unsupported CSI driver is already installed

jsafrane:
When opt-in CSI driver should be installed and the driver is already installed, but it's not the OpenShift one, Degrade the whole cluster and wait until user removes the other (community) CSI driver.

This affects only CSI drivers that are installed after TechPreviewNoUpgrade FeatureSet is set.

We distinguish "our" CSI driver by presence of annotation "csi.openshift.io/managed".

Passing CSIDriver to `shouldRunController` to simplify unit tests.

@openshift/storage

### DIFF
--- a/pkg/operator/csidriveroperator/driverstarter.go
+++ b/pkg/operator/csidriveroperator/driverstarter.go
@@ -2,6 +2,7 @@ package csidriveroperator
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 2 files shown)


jsafrane:
Tested on Azure, this is the Degraded condition on ClusterOperator `storage`:
```
  - lastTransitionTime: "2021-04-09T08:48:47Z"
    message: 'CSIDriverStarterDegraded: detected CSI driver disk.csi.azure.com that
      is not provided by OpenShift - please remove it before enabling the OpenShift
      one'
    reason: CSIDriverStarter_SyncError
    status: "True"
    type: Degraded
```

jsafrane:
/hold cancel
tested vsphere too

tsmetana (approved):

PR #153, opened 2021-04-08
Add vsphere syncer deployment

gnufied:
Ensure that vmware syncer is updated with latest image

I am going to put this on hold for a bit, because images aren't there yet.


bertinatto (commented):
Adding `lgtm` so we can just remove the `hold` tag when the syncer images are ready:

/lgtm

PR #167, opened 2021-05-17
Bug 1961120: added permissions to service monitoring

Elbehery:
This PR added permissions to operators service monitoring, allowing cluster version upgrade.

Currently the operators has `create` permission. When upgrading the cluster, it needs permission to `update` the service role.

### DIFF
--- a/assets/csidriveroperators/aws-ebs/03_role.yaml
+++ b/assets/csidriveroperators/aws-ebs/03_role.yaml
@@ -38,3 +38,6 @@ rules:
   verbs:
   - get
   - create
+  - update
+  - patch
+  - delete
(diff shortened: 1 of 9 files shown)


manifests/10_deployment-ibm-cloud-managed.yaml
```diff
@@ -1,4 +1,3 @@
-# *** AUTOMATICALLY GENERATED FILE - DO NOT EDIT ***
```
jsafrane, line 1: Why did you remove this comment?

Elbehery: I did not, I run `make update` after I changed the `yamls` 

btw, does the `yaml` need a new line at the end ?

jsafrane: `make verify` does not like this change.

Elbehery: Yes, this line was removed after executing `make update`. 
So I have `manually` restored the file, after executing `make update`, `make`, and `make test`

PR #173, opened 2021-06-02
Bug 1948090: Remove CSIDriverOperatorDeploymentAvailable condition when deploying CSI operator

bertinatto:
Currently, we deploy only one replica of the CSI operator Deployment. This apparently leads to CSO going unavailable when draining the node where a CSI operator is running on.

A reasonable way to fix this issue would be to run multiple replicas of the CSI operator, however, we chose to remove the _CSIDriverOperatorDeploymentAvailable_ condition instead. This condition was was being set to _False_ when draining the node where the CSI operator was running on, cascading the status up to the top _Available_ condition (through the unioning done in the CR controller in CSO).

From now on, the unioning that determines the top _Available_ state relies on the other _Available_ conditions reported by other controllers and by the CSI operator.

In addition to that, I introduced a health check for the CSI operator Deployment to make sure we go _Degraded_ if for some reason the CSI operator is not deployed successfully.

### DIFF
--- a/pkg/operator/csidriveroperator/deploymentcontroller.go
+++ b/pkg/operator/csidriveroperator/deploymentcontroller.go
@@ -6,12 +6,15 @@ import (
 	"strings"
 	"time"
 
+	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 
-	operatorapi "github.com/openshift/api/operator/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 	"github.com/openshift/library-go/pkg/operator/status"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
@@ -25,9 +28,12 @@ import (
 // It replace ${LOG_LEVEL} in the Deployment with current log level.
 // It replaces images in the Deployment using  CSIOperatorConfig.ImageReplacer.
 // It produces following Conditions:
-// <CSI driver name>CSIDriverOperatorDeploymentAvailable
 // <CSI driver name>CSIDriverOperatorDeploymentProgressing
 // <CSI driver name>CSIDriverOperatorDeploymentDegraded
+// This controller doesn't set the Available condition to avoid prematurely cascading
+// up to the clusteroperator CR a potential Available=false. On the other hand it
+// does a better in making sure the Degraded condition is properly set if the
+// Deployment isn't healthy.
 type CSIDriverOperatorDeploymentController struct {
 	name              string
 	operatorClient    v1helpers.OperatorClient
@@ -89,7 +95,7 @@ func (c *CSIDriverOperatorDeploymentController) Sync(ctx context.Context, syncCt
 	if err != nil {
 		return err
 	}
-	if opSpec.ManagementState != operatorapi.Managed {
+	if opSpec.ManagementState != operatorv1.Managed {
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


bertinatto:
/hold
I'm considering other options for this controller (other than setting multiple replicas).

pkg/operator/csidriveroperator/util.go
```diff
@@ -63,3 +68,46 @@ func reportUpdateEvent(recorder events.Recorder, obj runtime.Object, originalErr
...
+func checkDeploymentHealth(ctx context.Context, c appsclientv1.DeploymentsGetter, d *appsv1.Deployment) error {
+	d, err := c.Deployments(d.Namespace).Get(ctx, d.Name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	name := fmt.Sprintf("%s/%s", d.Namespace, d.Name)
+	if d.Generation > d.Status.ObservedGeneration {
+		return fmt.Errorf("deployment %s has a spec update not observed yet", name)
+	}
```
bertinatto, line 81: This might be overkill....

pkg/operator/csidriveroperator/deploymentcontroller.go
```diff
@@ -25,8 +27,6 @@ import (
 // It replace ${LOG_LEVEL} in the Deployment with current log level.
 // It replaces images in the Deployment using  CSIOperatorConfig.ImageReplacer.
 // It produces following Conditions:
-// <CSI driver name>CSIDriverOperatorDeploymentAvailable
-// <CSI driver name>CSIDriverOperatorDeploymentProgressing
```
bertinatto, line 29: @jsafrane currently CSO sets `Progressing=True` if it doesn't find any `*Progressing` condition:

https://github.com/bertinatto/cluster-storage-operator/blob/74c8159c8b6b1d616a4da3fdab1801e3e14f36c6/pkg/operator/csidriveroperator/crcontroller.go#L205-L213

jsafrane:
Progress of ClusterOperator `storage` installation without this PR: [before.txt](https://github.com/openshift/cluster-storage-operator/files/6624944/before.txt)

Progress of ClusterOperator `storage` installation with this PR: [after.txt](https://github.com/openshift/cluster-storage-operator/files/6624945/after.txt)

Notice that `CSIDriverOperatorDeploymentAvailable`/`CSIDriverOperatorDeploymentProgressing` is not used anywhere in `before.txt`, it seems that `AWSEBSCSIDriverOperatorCRAvailable` / `AWSEBSCSIDriverOperatorCRProgressing` is enough. From this it seems that this PR won't make the installation worse.

However, `after.txt` looks quite disturbed. I don't think it's because of this PR, but some things to notice there:

* At the beginning at 11:35:03 the ClusterOperator gets `Available=true` for 5 seconds and then gets `Available=false` because the CSI driver is not started yet. This happens probably because a default storage class is created (and its controller reports available=true) and before CSIDriverOperatorCRController syncs the ClusterCSIDriver for the first time (and there is no available=false for it yet). There are few conflicts reported in between, as the different CSO controller tries to update the storage CR or ClusterOperator, that might be the reason for 5 seconds.

* The cluster gets `Progressing=false` at 11:37:24. But then it becomes `Progressing=true` with:
    > AWSEBSCSIDriverOperatorCRProgressing: AWSEBSDriverNodeServiceControllerProgressing: Waiting for DaemonSet to deploy node pods

  I don't know what happened, could it be that worker nodes are added to the cluster? In that case Progressing=true could be actually correct.

jsafrane:
Checking cluster upgrade... with this PR I see a regression, `Progressing=true` is set later than it should be.

* Before, I can see the operator gets `Progressing=true` early with:
  > AWSEBSCSIDriverOperatorDeploymentProgressing: desired generation 2, current generation 1
  > AWSEBSCSIDriverOperatorDeploymentProgressing: 0 out of 1 pods running
  
  Then I can see `Progressing=false` for ~10 seconds, followed by `Progressing=true` with the rest of the update:
  > AWSEBSCSIDriverOperatorCRProgressing: AWSEBSDriverNodeServiceControllerProgressing: Waiting for DaemonSet to act on changes
  > AWSEBSCSIDriverOperatorCRProgressing: AWSEBSDriverControllerServiceControllerProgressing: Waiting for Deployment to deploy
   ...

* With this PR, I can see the operator is `Progressing=false` for quite same time, with messages in Degraded condition:
  > AWSEBSCSIDriverOperatorDeploymentDegraded: deployment openshift-cluster-csi-drivers/aws-ebs-csi-driver-operator has a spec update not observed yet
  
  Only after the operator update finishes I can see `Progressing=true` with expected
  > AWSEBSCSIDriverOperatorCRProgressing: AWSEBSDriverNodeServiceControllerProgressing: Waiting for DaemonSet to act on changes
  (the rest is the same as before the PR).

bertinatto:
/hold cancel
PR has been re-worked.

bertinatto:
> Checking cluster upgrade... with this PR I see a regression, `Progressing=true` is set later than it should be.
> 
>     * Before, I can see the operator gets `Progressing=true` early with:
>       > AWSEBSCSIDriverOperatorDeploymentProgressing: desired generation 2, current generation 1
>       > AWSEBSCSIDriverOperatorDeploymentProgressing: 0 out of 1 pods running
>       
>       
>       Then I can see `Progressing=false` for ~10 seconds, followed by `Progressing=true` with the rest of the update:
>       > AWSEBSCSIDriverOperatorCRProgressing: AWSEBSDriverNodeServiceControllerProgressing: Waiting for DaemonSet to act on changes
>       > AWSEBSCSIDriverOperatorCRProgressing: AWSEBSDriverControllerServiceControllerProgressing: Waiting for Deployment to deploy
>       > ...
>       
>       
>       This is wrong, but not related to this PR.
> 
>     * With this PR, I can see the operator is `Progressing=false` for quite same time during the upgrade, with messages in Degraded condition:
>       > AWSEBSCSIDriverOperatorDeploymentDegraded: deployment openshift-cluster-csi-drivers/aws-ebs-csi-driver-operator has a spec update not observed yet
>       
>       
>       So the whole operator Deployment update is tracked as `Progresing=false`, which is worse than before this PR. Only after the Deployment update finishes I can see `Progressing=true` with expected
>       > AWSEBSCSIDriverOperatorCRProgressing: AWSEBSDriverNodeServiceControllerProgressing: Waiting for DaemonSet to act on changes
>       > (the rest is the same as before the PR).

Added the progressing condition back and fixed the Deployment health checking.

pkg/operator/csidriveroperator/util.go
```diff
@@ -63,3 +68,42 @@ func reportUpdateEvent(recorder events.Recorder, obj runtime.Object, originalErr
 		recorder.Eventf(fmt.Sprintf("%sUpdated", gvk.Kind), "Updated %s:\n%s", resourcehelper.FormatResourceForCLIWithNamespace(obj), strings.Join(details, "\n"))
 	}
 }
+
+func checkDeploymentHealth(ctx context.Context, c appsclientv1.DeploymentsGetter, d *appsv1.Deployment) error {
+	d, err := c.Deployments(d.Namespace).Get(ctx, d.Name, metav1.GetOptions{})
```
bertinatto, line 73: Something very similar is done in CVO and OLM to make sure the Degraded condition is set when something is wrong with the operator Deployment.

jsafrane:
The code looks OK. I ran upgrade with this PR and I saw `storage` got new `status.version` very quickly and then it was flipping `Progressing: true` and `false` a lot.
See the attached file: [transitions.txt](https://github.com/openshift/cluster-storage-operator/files/6976959/transitions.txt)

`storage` was `Available: true` for the whole time.

I tried to do the same without this PR for comparison, but upgrade failed :-(.

jsafrane:
Nodes were drained & restarted during the update, but that should not drain / delete DaemonSet pods, I don't know what caused so much of `Waiting for DaemonSet to deploy node pods`.

bertinatto:
> Nodes were drained & restarted during the update, but that should not drain / delete DaemonSet pods, I don't know what caused so much of `Waiting for DaemonSet to deploy node pods`.

The `Progressing` condition is flipping because of  `AWSEBSCSIDriverOperatorCRProgressing`, which shouldn't be affected by this PR. You'll likely get the same results without this PR.

Regarding the `Waiting for DaemonSet to deploy node pods` message, that comes from the CSI operator and it was likely reported because of a node restart (when the node is back up, the expression `daemonSet.Status.NumberUnavailable > 0` becomes true, causing `AWSEBSCSIDriverOperatorCRProgressing == true` with the message above.

PR #175, opened 2021-06-07
Bug 1961317: Adding static assets to related objects

Elbehery:
This PR parses operator static assets, into operator related objects.

This PR depends on https://github.com/openshift/library-go/pull/1099


pkg/operator/csidriveroperator/driverstarter.go
```diff
@@ -56,8 +56,9 @@ type csiDriverControllerManager struct {
 	operatorConfig csioperatorclient.CSIOperatorConfig
 	// ControllerManager that installs the CSI driver operator and all its
 	// objects.
-	mgr     manager.ControllerManager
-	running bool
+	mgr                manager.ControllerManager
+	running            bool
+	ctrlRelatedObjects func() []configv1.ObjectReference
```
jsafrane, line 61: add `error` return to the function

pkg/operator/csidriveroperator/driverstarter.go
```diff
@@ -159,16 +163,27 @@ func (c *CSIDriverStarterController) sync(ctx context.Context, syncCtx factory.S
...
+	manager = manager.WithController(src, 1)
+	objs, err := src.RelatedObjects()
+	if err != nil {
+		// TODO : error msg or aggregate error
+		klog.Errorf("Can not find RelatedObjects, got %v", err)
+	} else {
+		ctrlRelatedObjects = func() []configv1.ObjectReference {
+			return objs
+		}
+	}
```
jsafrane, line 186: Please move `src.RelatedObjects` call to returned `func() ([]configv1.ObjectReference, error)` and call it during sync() - it can have proper error handling there and automatic retry on random API server error.

Elbehery: RelatedObject( ) can be only called from StaticResourceController .. Since we use type ControllerManager within csiDriverControllerManager , its not possible to call it within Sync()

Elbehery: actually thats why we added the funktor, so it has the related objects within the scope of `Sync()` and then we get the data from it, afterwards the funktor it self is `garbage` collected

jsafrane:
```
ctrlRelatedObjects = func() ([]configv1.ObjectReference, error) {
			return src.RelatedObjects()
		}
```

Elbehery: Ahh I c, I thought u want everything in Sync .. ok On it 👍🏽

Elbehery: @jsafrane Done 👍🏽

pkg/operator/csidriveroperator/driverstarter.go
```diff
@@ -56,8 +56,9 @@ type csiDriverControllerManager struct {
 	operatorConfig csioperatorclient.CSIOperatorConfig
 	// ControllerManager that installs the CSI driver operator and all its
 	// objects.
-	mgr     manager.ControllerManager
-	running bool
+	mgr                manager.ControllerManager
+	running            bool
+	ctrlRelatedObjects func() ([]configv1.ObjectReference, error)
```
deads2k, line 61: This looks worthy of an interface to me. I won't stick on the first one I see, but I'm likely to stick on it when we grow another.

Elbehery: +1 👍🏽

jsafrane: Please make an interface for this, the rest is lgtm

pkg/operator/csidriveroperator/driverstarter.go
```diff
@@ -152,7 +160,7 @@ func (c *CSIDriverStarterController) sync(ctx context.Context, syncCtx factory.S
 				Name:     ctrl.operatorConfig.CSIDriverName,
 			})
 			// add static assets
-			objs, err := ctrl.ctrlRelatedObjects()
+			objs, err := ctrl.getRelatedObjects()
```
jsafrane, line 163: Not sure if it's worth having separate `getRelatedObjects` function, you can call `ctrl.ctrlRelatedObjects,RelatedObjects()` directly here.

Elbehery: 👍🏽

PR #179, opened 2021-06-18
Use go:embed for static yaml files

jsafrane:
Move from bindata to go:embed.

I had to change some code to process error code from `ReadFile`, there is no `MustReadFile`.


jsafrane:
I don't see any storage related errors
/retest

PR #181, opened 2021-06-30
Add service account names to gcp credentials request manifest

akhil-rane:
For enabling short-lived credentials in gcp cluster using workload
identity (similar to AWS STS), we need ccoctl tool to know the
kubernetes service account names from the credentials request
manifest so that it can create gcp service accounts that can be
impersonated only by specific kubernetes service accounts.

ref: https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity


akhil-rane:
@jsafrane this looks good to merge

PR #182, opened 2021-06-30
Bug 1975538: Remove stale cruft installed by CVO in earlier releases

dobsonj:
See bug for details:
https://bugzilla.redhat.com/show_bug.cgi?id=1975538

### DIFF
--- a/manifests/03_delete_stale_objects.yaml
+++ b/manifests/03_delete_stale_objects.yaml
@@ -1,0 +1,159 @@
+# These 4 objects were added in 4.1 and removed in 4.6.
+# The release.openshift.io/delete annotation will remove them
+# from clusters that had them installed by previous releases.
+
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: cluster-storage-operator
+  namespace: openshift-cluster-storage-operator
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    release.openshift.io/delete: "true"
+rules:
+- apiGroups:
+  - storage.openshift.io
+  resources:
+  - "*"
+  verbs:
+  - "*"
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  - services
+  - endpoints
+  - persistentvolumeclaims
+  - events
+  - configmaps
+  - secrets
+  verbs:
+  - "*"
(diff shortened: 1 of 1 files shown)


manifests/03-cluster-role-binding.yaml
```diff
@@ -0,0 +1,6 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: cluster-storage-operator
+  annotations:
+    release.openshift.io/delete: "true"
```
bertinatto, line 6: Can we have these objects in a single file, say `1000-stale-cruft-whatever.yaml`?

dobsonj: Yeah, I think that's a good idea, I kind of debated whether or not to do that. I'll make that change.

manifests/03_delete_stale_objects.yaml
```diff
@@ -0,0 +1,147 @@
...
+# The release.openshift.io/delete annotation will remove them
+# from clusters that had them installed by previous releases.
+
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: cluster-storage-operator
+  namespace: openshift-cluster-storage-operator
+  annotations:
+    release.openshift.io/delete: "true"
```
wking, line 11: You probably also need the cluster-profile annotations like:

```yaml
    include.release.openshift.io/ibm-cloud-managed: "true"
```

otherwise the CVO will completely ignore the manifest.

dobsonj: Thank you! I was scratching my head why the manifest wasn't getting applied during testing.

dobsonj: @wking After adding those cluster-profile annotations, the manifest is getting applied. However... the objects are not getting deleted. I'm wondering what I'm missing.
I tested 2 scenarios:
1) fresh install of 4.7.18 (does not include the original manifests), then upgrade to a release with these changes.
2) apply the original manifests to a new 4.7.18 cluster, then upgrade to a release with these changes.
In case 1, the objects got created with the delete annotation. In case 2, the delete annotation was added to the existing objects. But in both cases, those objects were never deleted. Is this expected? What actually triggers CVO to delete objects based on the annotation?

dobsonj:
/close
I need to re-test this on a more recent build and see if the objects are still not getting cleaned up as described in https://github.com/openshift/cluster-storage-operator/pull/182#discussion_r665715504

PR #184, opened 2021-07-08
exclude default storage class on Azure StackHub

tjungblu:
this PR makes sure we don't create an incompatible storageclass on ASH which uses the CSI driver.

follow-up from https://github.com/openshift/azure-disk-csi-driver-operator/pull/26


jsafrane:
/lgtm
/approve
CSI driver installation works, I can see only one test flaked in the last attempt.

PR #187, opened 2021-07-19
Support external control plane topology

csrwng:
If running on a cluster with control plane topology of type 'External' (as in ROKS or Hypershift), CSI operators should not be scheduled to masters. This change sets the node selector to empty if this is the case.

### DIFF
--- a/pkg/operator/csidriveroperator/csioperatorclient/types.go
+++ b/pkg/operator/csidriveroperator/csioperatorclient/types.go
@@ -41,6 +41,8 @@ type CSIOperatorConfig struct {
 	OLMOptions *OLMOptions
 	// Run the CSI driver operator only when given FeatureGate is enabled
 	RequireFeatureGate string
+	// Schedule CSI driver operator on workers when control plane is externalized
+	ScheduleOnWorkers bool
 }
 
 // OLMOptions contains information that is necessary to remove old CSI driver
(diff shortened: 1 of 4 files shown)


pkg/operator/csidriveroperator/csioperatorclient/types.go
```diff
@@ -41,6 +41,8 @@ type CSIOperatorConfig struct {
 	OLMOptions *OLMOptions
 	// Run the CSI driver operator only when given FeatureGate is enabled
 	RequireFeatureGate string
+	// Schedule CSI driver operator on workers when control plane is externalized
+	ScheduleOnWorkers bool
```
bertinatto, line 45: @csrwng on vSphere clusters we also schedule the vsphere-problem-detector pod on master nodes, is that something we need to change as well?

csrwng: Yes, thank you, we should. Will update this PR

jsafrane:
Is there a generic document describing what runs where when using various deployment tools like installer (with masters), hypershift (without masters) etc? These PRs are nice, however, there are many things that need to be changed in all storage operators.

bertinatto (commented):
/lgtm
The flake seems unrelated to this PR.

PR #188, opened 2021-07-21
Add new driver permissions for vmware-vsphere-csi-operator

bertinatto:
CC @openshift/storage

### DIFF
--- a/assets/csidriveroperators/vsphere/06_clusterrole.yaml
+++ b/assets/csidriveroperators/vsphere/06_clusterrole.yaml
@@ -276,18 +276,13 @@ rules:
   resources:
   - customresourcedefinitions
   verbs:
-  - get
+  - '*'
 - apiGroups:
   - cns.vmware.com
   resources:
-  - cnsvspherevolumemigrations
+  - '*'
   verbs:
-  - create
-  - get
-  - list
-  - watch
-  - update
-  - delete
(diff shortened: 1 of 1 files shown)


bertinatto:
/hold
until I manually test this

bertinatto:
/hold cancel
I'm unable to test this at the moment because I wasn't able to get a working cluster.

assets/csidriveroperators/vsphere/06_clusterrole.yaml
```diff
@@ -276,18 +276,13 @@ rules:
...
-  - cnsvspherevolumemigrations
+  - '*'
   verbs:
-  - create
-  - get
-  - list
-  - watch
-  - update
-  - delete
+  - '*'
```
gnufied, line 285: This is for enabling webhook right?

bertinatto: Correct, to make sure the webhook can start we need to enable the CSI migration in the driver; and once we do that the driver starts meddling with objects from this api group.

bertinatto:
It seems like previous failures in `e2e-aws-upgrade` job fixed by https://github.com/openshift/installer/pull/5100, however, recent failures are caused by fired alerts that aren't related to this PR.

PR #197, opened 2021-08-02
 Bug 1986557: Install CSI drivers on all platforms

adambkaplan:
Allow platform-agnostic CSI drivers to be installed on all platforms.
This allows CSI drivers that do not depend on a specific cloud provider
to be installed by the cluster storage operator.

### DIFF
--- a/pkg/operator/csidriveroperator/csioperatorclient/types.go
+++ b/pkg/operator/csidriveroperator/csioperatorclient/types.go
@@ -8,6 +8,12 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+const (
+	// AllPlatforms is a special PlatformType that indicates a CSI driver is installable on any cloud provider.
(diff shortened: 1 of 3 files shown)


sbose78:
👍

coreydaley:
/retest
```
 level=info msg=Creating infrastructure resources...
level=error
level=error msg=Error: Provider produced inconsistent result after apply
level=error
level=error msg=When applying changes to module.vpc.aws_route_table.private_routes[1],
level=error msg=provider "registry.terraform.io/-/aws" produced an unexpected new value for
level=error msg=was present, but now absent.
level=error
level=error msg=This is a bug in the provider, which should be reported in the provider's own
level=error msg=issue tracker.
level=error
level=fatal msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to apply Terraform: failed to complete the change 
```

PR #198, opened 2021-08-03
[WIP] Jira build 284 integrate shared resources operator

coreydaley:
https://issues.redhat.com/browse/BUILD-284


manifests/image-references
```diff
@@ -102,3 +102,11 @@ spec:
...
       kind: DockerImage
       name: quay.io/openshift/origin-kube-rbac-proxy:latest
+  - name: csi-driver-shared-resource-operator
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-csi-driver-shared-resource-operator:latest
+  - name: csi-driver-shared-resource
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-csi-driver-shared-resource:latest
```
adambkaplan, line 112: /hold

We can't add these image references until we have the ART pipeline set up and have an image built in OSBS. Otherwise we break the payload and block nightly releases.

assets/csidriveroperators/shared-resource/09_daemonset.yaml
```diff
@@ -0,0 +1,59 @@
+apiVersion: apps/v1
+kind: Daemonset
```
jsafrane, line 2: We usually don't deploy *operators* as DaemonSets. Please use a Deployment and copy some recent one - we added support for SNO (that should run 1 replica only) and for hypershift (which does not have master nodes).

assets/csidriveroperators/shared-resource/07_role_shared_resource_config.yaml
```diff
@@ -0,0 +1,15 @@
+# Allow AWS EBS CSI driver operator to read CA bundle from openshift-config-managed/kube-cloud-config ConfigMap
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: shared-resource-csi-driver-operator-aws-config-role
```
jsafrane, line 5: I think the operator is not AWS specific, please remove aws (and maybe aws-config) from the names.

pkg/operator/csidriveroperator/csioperatorclient/shared-resource.go
```diff
@@ -0,0 +1,41 @@
...
+func GetSharedResourceCSIOperatorConfig() CSIOperatorConfig {
+	pairs := []string{
+		"${OPERATOR_IMAGE}", os.Getenv(envSharedResourceDriverOperatorImage),
+		"${DRIVER_IMAGE}", os.Getenv(envSharedResourceDriverImage),
+	}
+
+	return CSIOperatorConfig{
+		CSIDriverName:   SharedResourceDriverName,
+		ConditionPrefix: "SHARES",
+		Platform:        configv1.AWSPlatformType,
```
jsafrane, line 25: this should be a new "wildcard" platform.

pkg/operator/csidriveroperator/csioperatorclient/shared-resource.go
```diff
@@ -0,0 +1,41 @@
...
+		StaticAssets: []string{
+			"csidriveroperators/shared-resource/02_sa.yaml",
+			"csidriveroperators/shared-resource/03_role.yaml",
+			"csidriveroperators/shared-resource/04_rolebinding.yaml",
+			"csidriveroperators/shared-resource/05_clusterrole.yaml",
+			"csidriveroperators/shared-resource/06_clusterrolebinding.yaml",
+			"csidriveroperators/shared-resource/07_role_aws_config.yaml",
+			"csidriveroperators/shared-resource/08_rolebinding_aws_config.yaml",
```
jsafrane, line 33: please remove aws from the file names

assets/csidriveroperators/shared-resource/05_clusterrole.yaml
```diff
@@ -0,0 +1,269 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
jsafrane, line 2: Does the CSI driver need access to some CR for the driver? I don't see its RBAC here. In addition, who is creating the CRD? It could be CVO (then put the CRD into `/manifests/`) or it could be your operator (then the operator needs permissions to create/update CRDs).

jsafrane:
Out of curiosity, do you aim at OCP 4.9? We're past feature freeze there and this looks like a feature.

adambkaplan:
@jsafrane this now being staged for when 4.10 is available. We missed ART's cutoff for adding things to the payload.

jsafrane:
I was assured ART builds are present.
/hold cancel

PR #200, opened 2021-08-05
Bug 1990601: Always run CSI driver controller for Azure Stack Hub

Fedosin:
Since Azure Stack Hub does not support in-tree volume plugin for azure disk, we have to always enable the CSI driver for this platform.

This commit checks if we use Azure Stack Hub and enables the driver, even if the feature gate may not be set.


jsafrane:
/lgtm
/approve
Thanks for the PR!

PR #209, opened 2021-09-02
Bug 1999674: Update prometheus rule to alert about esxi host versions

gnufied:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1999674


jsafrane:
/override ci/prow/e2e-azure-csi
/override ci/prow/e2e-gcp

They can't be broken by this PR.

PR #211, opened 2021-09-03
Bug 1989740: Move manila operator role under the manila namespace

mandre:
Manila operands doesn't have access to the
`openshift-cluster-csi-drivers` namespace under which the operator is
deployed. We need to move the role to the `openshift-manila-csi-driver`
namespace.

This has worked until now because the rules are also covered in the
ClusterRole.


mandre:
We need to review the ClusterRole for unnecessary rules.

mandre:
/remove-lifecycle stale
I'll get back to it at some point.

PR #215, opened 2021-09-20
Bug 2001761: Fix RelatedObjects when RBAC API is missing

jsafrane:
When starting a CSI driver operator and discovering its dependencies / `RelatedObjects`, some APIs may be temporarily missing in the API server. RESTMapper is caching the error responses, so make sure to refresh the cache in the next sync.

I tried to fix the caching in [client-go](https://github.com/kubernetes/kubernetes/pull/104814), but it seems it's not going anywhere.

### DIFF
--- a/pkg/csoclients/csoclients.go
+++ b/pkg/csoclients/csoclients.go
@@ -1,11 +1,15 @@
 package csoclients
 
 import (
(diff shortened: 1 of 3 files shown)


jsafrane:
cc @openshift/storage

PR #218, opened 2021-10-01
podsecurity: enforce privileged for openshift-cluster-csi-drivers namespace

s-urbaniak:
Starting with OpenShift 4.10 we are introducing PodSecurity admission (https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/2579-psp-replacement).

Currently, all pods are marked as privileged, however, over time we want to enforce at least baseline, admirably restricted as default. In order not to break control plane workloads this allows workloads in `openshift-cluster-csi-drivers` namespace to run privileged pods.

See https://github.com/openshift/enhancements/pull/899 for more details (and excuse the eventual consistency of updates).

/cc @stlaz

### DIFF
--- a/manifests/02_csi_driver_operators_namespace.yaml
+++ b/manifests/02_csi_driver_operators_namespace.yaml
@@ -10,3 +10,6 @@ metadata:
     workload.openshift.io/allowed: "management"
   labels:
     openshift.io/cluster-monitoring: "true"
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/warn: privileged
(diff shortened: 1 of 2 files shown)


manifests/02_csi_driver_operators_namespace.yaml
```diff
@@ -10,3 +10,6 @@ metadata:
     workload.openshift.io/allowed: "management"
   labels:
     openshift.io/cluster-monitoring: "true"
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/warn: privileged
```
bertinatto, line 15: For historical reasons, the Manila CSI driver is deployed to a different namespace than the other CSI drivers (_assets/csidriveroperators/manila/01_namespace.yaml_). That driver also has a component that needs to be privileged. 

Could you add these labels to that namespace as well? Note that that NS is created by this operator, not CVO.

s-urbaniak: @bertinatto added the labels, thank you!

bertinatto (commented):
/lgtm
/approve

PR #224, opened 2021-10-15
bump openshift/api to get csi.sharedresource.openshift.io

coreydaley:


coreydaley:
/assign @jsafrane 
We need to pull in this update for https://github.com/openshift/csi-driver-shared-resource-operator/pull/9 to pass, since it is rejecting the `csi.sharedresource.openshift.io` driver name.

PR #229, opened 2021-10-25
Update Shared Resource Operator permissions

coreydaley:
Reign in the Shared Resource Operator permissions
for Custom Resource Definitions.

We need a blanket CREATE permission for ALL CRDs
But only need list,get,watch,delete for sharedconfigmaps and
sharedsecrets.

### DIFF
--- a/assets/csidriveroperators/shared-resource/05_clusterrole.yaml
+++ b/assets/csidriveroperators/shared-resource/05_clusterrole.yaml
@@ -88,9 +88,20 @@ rules:
   resources:
   - customresourcedefinitions
   verbs:
-  - list
   - create
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  resourceNames:
+  - sharedconfigmaps.sharedresource.openshift.io
+  - sharedsecrets.sharedresource.openshift.io
+  verbs:
+  - get
+  - list
   - watch
+  - update
+  - patch
   - delete
 - apiGroups:
   - coordination.k8s.io
@@ -110,6 +121,7 @@ rules:
   - secrets
   - configmaps
   verbs:
+  - create
   - get
   - list
   - watch


assets/csidriveroperators/shared-resource/05_clusterrole.yaml
```diff
@@ -88,8 +88,15 @@ rules:
   resources:
   - customresourcedefinitions
   verbs:
-  - list
```
adambkaplan, line 91: `list`,`get`, and `watch` are IMO safe to grant for all CRDs.

Granting list to all CRDs is needed to avoid issues with listers - a serviceaccount with restrictions on list needs to provide the name in a field selector, which is cumbersome.

See https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources

assets/csidriveroperators/shared-resource/05_clusterrole.yaml
```diff
@@ -88,8 +88,15 @@ rules:
...
   - create
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions/sharedconfigmaps.sharedresource.openshift.io
+  - customresourcedefinitions/sharedsecrets.sharedresource.openshift.io
+  verbs:
+  - list
+  - get
```
adambkaplan, line 99: This should use `resourceNames` in the RBAC, and restrict `update`, and `delete`:

```suggestion
- apiGroups:
  - apiextensions.k8s.io
  resources:
  - customresourcedefinitions
  resourceNames:
  - sharedconfigmaps.sharedresource.openshift.io
  - sharedsecrets.sharedresource.openshift.io
  verbs:
  - update
  - delete
```

adambkaplan (changes requested):

assets/csidriveroperators/shared-resource/05_clusterrole.yaml
coreydaley: We are using the fieldSelector (https://github.com/openshift/csi-driver-shared-resource-operator/blob/master/pkg/operator/starter.go#L88-L90), and honestly I think we are being slack in our permissions granting (probably throughout openshift honestly) to save a line of code.

adambkaplan: > We are using the fieldSelector (https://github.com/openshift/csi-driver-shared-resource-operator/blob/master/pkg/operator/starter.go#L88-L90), and honestly I think we are being slack in our permissions granting (probably throughout openshift honestly) to save a line of code.

ack, then :+1:

assets/csidriveroperators/shared-resource/05_clusterrole.yaml
```diff
@@ -88,8 +88,18 @@ rules:
...
+  - customresourcedefinitions
+  resourceNames:
+  - sharedconfigmaps.sharedresource.openshift.io
+  - sharedsecrets.sharedresource.openshift.io
+  verbs:
+  - get
+  - update
+  - list
   - watch
   - delete
```
adambkaplan, line 104: Apologies, forgot to add `patch` to the list of verbs here. I tend to put the verbs in Read -> Create -> Update -> Delete ordering:

```suggestion
  verbs:
  - get
  - list
  - watch
  - update
  - patch
  - delete
```

adambkaplan (changes requested):

adambkaplan (approved):
/lgtm

gabemontero:
hmm ... ideally @coreydaley we would not have merged this without running the new e2e-aws-shared-resources test 

we'll have to consider these changes as well when investigating why the test is failing in https://github.com/openshift/cluster-storage-operator/pull/231

gabemontero:
> hmm ... ideally @coreydaley we would not have merged this without running the new e2e-aws-shared-resources test
> 
> we'll have to consider these changes as well when investigating why the test is failing in #231

I have throw away PR https://github.com/openshift/cluster-storage-operator/pull/232 up to try out e2e-aws-shared-resources against effectively the master branch

gabemontero:
> > hmm ... ideally @coreydaley we would not have merged this without running the new e2e-aws-shared-resources test
> > we'll have to consider these changes as well when investigating why the test is failing in #231
> 
> I have throw away PR #232 up to try out e2e-aws-shared-resources against effectively the master branch

the e2e-aws-shared-resource run from https://github.com/openshift/cluster-storage-operator/pull/232 look good

so no problems with this PR

PR #234, opened 2021-11-04
WIP: BUILD-256: add shared resource driver configuration configmap

gabemontero:
proved we could not use library-go static resource controller in shared resource operator;
it is truly "static" ... if we change the data field, the controller changes it back

but putting it in assets leverages the static resource controller as well in this repo (thanks for the reminder @coreydaley )

creating  WIP PR with putting it in the CSO manifest, so the CVO creates it, similar to our other CVO based controllers

will launch cluster bot cluster to test if we can change 

otherwise, we are back to the SRO either "manually" creating the configmap outside of library-go, or having to employ configobservers to deal with it

### DIFF
--- a/manifests/13_config_configmap.yaml
+++ b/manifests/13_config_configmap.yaml
@@ -1,0 +1,37 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: csi-driver-shared-resource-config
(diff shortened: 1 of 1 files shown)


gabemontero:
yeah this didn't work ... not sure why @coreydaley 

but a clusterbot cluster with this did *not* create the configmap

PR #238, opened 2021-11-22
Add CredentialsRequest for Alibaba Disk CSI driver

jsafrane:
cc @openshift/storage 

Based on an example in https://github.com/openshift/cloud-credential-operator/pull/412 and initial list of permissions I got from Alibaba on slack .

### DIFF
--- a/manifests/03_credentials_request_alibaba.yaml
+++ b/manifests/03_credentials_request_alibaba.yaml
@@ -1,0 +1,50 @@
+apiVersion: cloudcredential.openshift.io/v1
+kind: CredentialsRequest
+metadata:
+  name: alibaba-disk-csi-driver-operator
+  namespace: openshift-cloud-credential-operator
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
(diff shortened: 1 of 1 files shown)


manifests/03_credentials_request_alibaba.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - ecs:ModifyDiskSpec
+      - ecs:CreateSnapshot
+      - ecs:DescribeSnapshotGroups
+      - ecs:CreateSnapshotGroup
+      - ecs:DeleteSnapshotGroup
+      - ecs:CopySnapshot
+      - ecs:DeleteDisk
+      - ecs:DescribeInstanceAttribute
+      - ecs:DescribeInstances
+      - cs:CreateResourcesSystemTags
```
bertinatto, line 44: Should this be `ecs:`? I couldn't find this action in the official docs.

jsafrane: I asked Aliababa and they said `cs:` is correct

PR #239, opened 2021-11-23
Add Alibaba Disk CSI driver

jsafrane:
Tested with force-installing the driver on AWS, it needs https://github.com/openshift/alibaba-disk-csi-driver-operator/pull/12, but otherwise it started the CSI driver DaemonSet + Deployment. Driver pods crashed because they were running on a wrong cloud.


jsafrane:
cc @openshift/storage

bertinatto (commented):
/lgtm

PR #240, opened 2021-11-23
Replace registry.ci images with quay

jsafrane:
The images are already mirrored to quay.io.

cc @openshift/storage

PR #241, opened 2021-11-30
Bug 2027363: Fix reference to Azure File images

jsafrane:
Broken in https://github.com/openshift/cluster-storage-operator/pull/240

@openshift/storage


bertinatto (commented):
/lgtm

PR #242, opened 2021-12-06
Add IBM VPC Block CSI Driver Operator

dobsonj:
Epic: https://issues.redhat.com/browse/STOR-598
I was able to get the driver and operator deployed with these changes and verify that PVC provisioning works:
```
[jon@dagobah ibm]$ oc get pods -n openshift-cluster-csi-drivers
NAME                                                 READY   STATUS    RESTARTS   AGE
ibm-vpc-block-csi-controller-5548dcc7b7-wc9z4        5/5     Running   0          3m10s
ibm-vpc-block-csi-driver-operator-6cdd5d465f-nsrlg   1/1     Running   0          3m16s
ibm-vpc-block-csi-node-4bp86                         3/3     Running   0          2m42s
ibm-vpc-block-csi-node-bgrjn                         3/3     Running   0          3m9s
ibm-vpc-block-csi-node-fxvvk                         3/3     Running   0          2m27s
ibm-vpc-block-csi-node-g57sz                         3/3     Running   0          2m12s
ibm-vpc-block-csi-node-nrsnb                         3/3     Running   0          119s
ibm-vpc-block-csi-node-zsrhw                         3/3     Running   0          2m57s
[jon@dagobah ibm]$ oc create -f validPVC.yaml
persistentvolumeclaim/csi-block-pvc-good created
[jon@dagobah ibm]$ oc create -f validPOD.yaml
pod/ubuntu created
[jon@dagobah ibm]$ oc get pods
NAME     READY   STATUS    RESTARTS   AGE
ubuntu   1/1     Running   0          47s
[jon@dagobah ibm]$ oc get pvc
NAME                 STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS                AGE
csi-block-pvc-good   Bound    pvc-02efd13d-a3f1-4e60-bd79-5625ec3eeed0   10Gi       RWO            ibmc-vpc-block-5iops-tier   67s
[jon@dagobah ibm]$ oc get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM                        STORAGECLASS                REASON   AGE
pvc-02efd13d-a3f1-4e60-bd79-5625ec3eeed0   10Gi       RWO            Delete           Bound    default/csi-block-pvc-good   ibmc-vpc-block-5iops-tier            48s
```
/cc @openshift/storage

PR #243, opened 2021-12-08
Bug 2030364: Setup port, mount TLS cert into CSI shared resource operator

alicerum:
Currently CSI shared resource operator pod does not know about the TLS certificate it has to use and the port for the metrics. That should be changed in the deployment.

### DIFF
--- a/assets/csidriveroperators/shared-resource/09_deployment.yaml
+++ b/assets/csidriveroperators/shared-resource/09_deployment.yaml
@@ -36,6 +36,13 @@ spec:
           requests:
             memory: 50Mi
             cpu: 10m
+        ports:
+        - name: provisioner-m
+          containerPort: 6000
+          protocol: TCP
+        volumeMounts:
+        - mountPath: /etc/secrets
+          name: shared-resource-csi-driver-operator-metrics-serving-cert
       priorityClassName: system-cluster-critical
       serviceAccountName: shared-resource-csi-driver-operator
       nodeSelector:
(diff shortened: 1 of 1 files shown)


jsafrane:
@alicerum are you sure this PR enables `openshift_csi_share_configmap_total` metrics? Is this metric emitted by the operator? You should probably expose ports of the CSI driver, not the CSI driver operator.

alicerum:
@jsafrane yes, this metric (amount of shared resources gauge) is produced by the operator. We don't produce it in the driver because driver is deployed in a `DaemonSet`, so we don't want all the nodes contribute to the same metric.
CSI driver produces the shared resource mounts attempt counter metric.

gabemontero:
ok the vsphere-csi e2e got passed flakes and passed @jsafrane 

can we skip the remaining ones, or do we need to keep on retesting ?

and to +1 @alicerum 's prior response to you, yes, per original recommendation form @adambkaplan given the reason's @alicerum noted we have metrics in both the shared resource operator and shared resource driver.  This is intentional.

jsafrane:
All "Required" jobs must pass and they did. azure-file-csi and vsphere are optional.

gabemontero:
a dependency related curl failed during the verify job

/test verify

gabemontero:
hmmm ... can't ge the verify job with flaked on the curl download to refire ... might have to close/reopen this PR

will try 

/refresh

and wait for the upgrade job to finish

dobsonj:
/retest-required
PR #242 just merged with the fix for `make verify`

PR #253, opened 2022-01-14
no CredentialsRequests in ibm-cloud-managed

joelddiaz:
The ibm-cloud-managed profile doesn't use the cloud-credential-operator.
The CredentialsRequest CRs should not be installed in that environment.

Unmark the existing CredentialsRequests so they are no longer installed
in ibm-cloud-managed.

Create a list of previously-installed CredentialsRequests so that they
are cleaned up by the CVO only on the ibm-cloud-managed profile.

xref: https://issues.redhat.com/browse/CCO-177

### DIFF
--- a/manifests/03_credentials_request_alibaba.yaml
+++ b/manifests/03_credentials_request_alibaba.yaml
@@ -4,7 +4,6 @@ metadata:
   name: alibaba-disk-csi-driver-operator
   namespace: openshift-cloud-credential-operator
   annotations:
-    include.release.openshift.io/ibm-cloud-managed: "true"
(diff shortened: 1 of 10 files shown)


jsafrane:
@joelddiaz do you have a BZ for this or can it wait for 4.11?

joelddiaz:
@csrwng Am I right that this ibm-cloud-managed cleanup work can wait for 4.11?

csrwng:
@joelddiaz yes it can

jsafrane:
/hold
for 4.11 to open. Just to silence openshift-bot.

PR #254, opened 2022-01-14
Bug 2040880: Do not replace unknown condition with true when we union conditions

gnufied:
xref https://bugzilla.redhat.com/show_bug.cgi?id=2040880

### DIFF
--- a/pkg/operator/csidriveroperator/crcontroller.go
+++ b/pkg/operator/csidriveroperator/crcontroller.go
@@ -216,10 +216,15 @@ func (c *CSIDriverOperatorCRController) syncConditions(ctx context.Context, cond
 		}
 	}
 
-	upgradeableCond := status.UnionCondition(operatorapi.OperatorStatusTypeUpgradeable, operatorapi.ConditionTrue, nil, conditions...)
-	upgradeableCond.Type = c.crConditionName(operatorapi.OperatorStatusTypeUpgradeable)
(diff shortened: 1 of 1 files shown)


pkg/operator/csidriveroperator/crcontroller.go
```diff
@@ -217,10 +217,6 @@ func (c *CSIDriverOperatorCRController) syncConditions(ctx context.Context, cond
 	}
 
 	upgradeableCond := status.UnionCondition(operatorapi.OperatorStatusTypeUpgradeable, operatorapi.ConditionTrue, nil, conditions...)
-	upgradeableCond.Type = c.crConditionName(operatorapi.OperatorStatusTypeUpgradeable)
-	if upgradeableCond.Status == operatorapi.ConditionUnknown {
-		upgradeableCond.Status = operatorapi.ConditionTrue
-	}
```
jsafrane, line 223: I think it is correct. At least it worked for AWS. With this PR, OCP can't be installed:

```
Some cluster operators are not ready: storage (Upgradeable=Unknown _NoData: )
```

gnufied: hmm, I think this is because the Union code in library-go adds "Condition: Unknown" if that condition type is not there in the CR's condition at all. May be we should only union if `Condition: Unknown` is present.

PR #255, opened 2022-01-17
Bug 2037856: Update library-go to get leader election updates

jsafrane:
Get ConfigMap + Lease leader election in 4.10 from https://github.com/openshift/library-go/pull/1282.

And bump go version to 1.17 when at it.

@openshift/storage


gnufied:
The CSI failures look unrelated.

/lgtm

PR #259, opened 2022-01-27
Bug 2047025: Add "patch" permissions to Alibaba CSI driver operator

jsafrane:
They're required by the new external-snapshotter sidecar.

cc @openshift/storage


bertinatto (commented):
/lgtm

PR #266, opened 2022-03-31
Azure File CSI Driver Operator is GA in OCP 4.11

bertinatto:
Epic: https://issues.redhat.com/browse/STOR-736

CC @openshift/storage


bertinatto:
/hold
for manual testing. Since the operator and the driver will be installed by default, we need to make sure it won't break Azure clusters.

bertinatto:
/hold cancel

Spun up a cluster and it seems to be working OK. The Azure File operator was installed correctly and I was able to dynamically-provision a volume.

I also checked a CI run and it seems OK as well.

PR #267, opened 2022-03-31
Add missing ibm cloud annotations to prometheus rbac

alvaroaleman:
Currently, the servicemonitor but not the rbac is deployed which breaks
the scraping.


jmcmeek:
@alvaroaleman An IBM customer reported this on an earlier release. Can this be cherry-picked to earlier releases?  It looks like the affected manifest appears in 4.7.

PR #273, opened 2022-04-27
Bug 2077050: OCP should default to pd-ssd disk type on GCP

RomanBednar:
Merge along with [PR](https://github.com/openshift/gcp-pd-csi-driver-operator/pull/47)

cc @openshift/storage


jsafrane:
I think we should not change the storage class name. After upgrade, both `standard` and `ssd` will exist and IMO both will be default.

RomanBednar:
> I think we should not change the storage class name. After upgrade, both `standard` and `ssd` will exist and IMO both will be default.

I've changed the patch according to our discussion and BZ comment. Same for the operator PR (linked in the description).

PR #277, opened 2022-05-09
Bug 2081557: Fix DefaultStorageClassController getting Available=False on error

jsafrane:
Don't flip `DefaultStorageClassController` `Available=true` -> `false` when it fails to talk to the API server. Storage is OK, most probably the API server is being updated.

When the controller fails constantly, it will get `Degraded=true` with some inertia.


bertinatto (commented):
/lgtm

PR #278, opened 2022-05-09
Bug 2086231: BUILD-405: Install the Shared Resource CSI Driver WebHook

jkhelil:
…rce csi driver

### DIFF
--- a/assets/csidriveroperators/shared-resource/09_deployment.yaml
+++ b/assets/csidriveroperators/shared-resource/09_deployment.yaml
@@ -25,6 +25,8 @@ spec:
           value: ${DRIVER_IMAGE}
         - name: NODE_DRIVER_REGISTRAR_IMAGE
           value: ${NODE_DRIVER_REGISTRAR_IMAGE}
+        - name: WEBHOOK_IMAGE
+          value: ${WEBHOOK_IMAGE}
         - name: POD_NAME
           valueFrom:
             fieldRef:
--- a/manifests/image-references
+++ b/manifests/image-references
@@ -122,6 +122,10 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-csi-driver-shared-resource:latest
+  - name: csi-driver-shared-resource-webhook
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-csi-driver-shared-resource-webhook:latest
   - name: alibaba-disk-csi-driver-operator
     from:
       kind: DockerImage
--- a/pkg/operator/csidriveroperator/csioperatorclient/shared-resource.go
+++ b/pkg/operator/csidriveroperator/csioperatorclient/shared-resource.go
@@ -9,12 +9,14 @@ const (
 	SharedResourceDriverName             = "csi.sharedresource.openshift.io"
 	envSharedResourceDriverOperatorImage = "SHARED_RESOURCE_DRIVER_OPERATOR_IMAGE"
 	envSharedResourceDriverImage         = "SHARED_RESOURCE_DRIVER_IMAGE"
+	envSharedResourceDriverWebhookImage  = "SHARED_RESOURCE_DRIVER_WEBHOOK_IMAGE"
 )
 
 func GetSharedResourceCSIOperatorConfig() CSIOperatorConfig {
 	pairs := []string{
 		"${OPERATOR_IMAGE}", os.Getenv(envSharedResourceDriverOperatorImage),
 		"${DRIVER_IMAGE}", os.Getenv(envSharedResourceDriverImage),
+		"${WEBHOOK_IMAGE}", os.Getenv(envSharedResourceDriverWebhookImage),
 	}
 
 	return CSIOperatorConfig{
--- a/manifests/10_deployment-ibm-cloud-managed.yaml
+++ b/manifests/10_deployment-ibm-cloud-managed.yaml
@@ -85,6 +85,8 @@ spec:
           value: quay.io/openshift/origin-csi-driver-shared-resource-operator:latest
         - name: SHARED_RESOURCE_DRIVER_IMAGE
           value: quay.io/openshift/origin-csi-driver-shared-resource:latest
+        - name: SHARED_RESOURCE_DRIVER_WEBHOOK_IMAGE
+          value: quay.io/openshift/origin-csi-driver-shared-resource-webhook:latest
         - name: ALIBABA_DISK_DRIVER_OPERATOR_IMAGE
           value: quay.io/openshift/origin-alibaba-disk-csi-driver-operator:latest
         - name: ALIBABA_CLOUD_DRIVER_IMAGE
--- a/manifests/10_deployment.yaml
+++ b/manifests/10_deployment.yaml
@@ -112,6 +112,8 @@ spec:
             value: quay.io/openshift/origin-csi-driver-shared-resource-operator:latest
           - name: SHARED_RESOURCE_DRIVER_IMAGE
             value: quay.io/openshift/origin-csi-driver-shared-resource:latest
+          - name: SHARED_RESOURCE_DRIVER_WEBHOOK_IMAGE
+            value: quay.io/openshift/origin-csi-driver-shared-resource-webhook:latest
           - name: ALIBABA_DISK_DRIVER_OPERATOR_IMAGE
             value: quay.io/openshift/origin-alibaba-disk-csi-driver-operator:latest
           - name: ALIBABA_CLOUD_DRIVER_IMAGE


gabemontero:
@jkhelil can you preface your PR title with the Jira this is for

assets/csidriveroperators/shared-resource/09_deployment.yaml
```diff
@@ -17,6 +17,18 @@ spec:
         name: shared-resource-csi-driver-operator
     spec:
       containers:
+      - image: ${WEBHOOK_IMAGE}
```
gabemontero, line 20: Typically the webhook would be in its own deployment @jkhelil, not a separate container in the operator's deployment.

I believe I missed planning for this.  Were you given direction there by @coreydaley or @adambkaplan to do it this way?

I would be surprised if that was the case.  If so, let's discuss.  If not though, adjust this PR to create a separate deployment.

adambkaplan: +1 - the webhook should be in a separate deployment.

manifests/image-references
```diff
@@ -122,6 +122,10 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-csi-driver-shared-resource:latest
+  - name: csi-driver-shared-resource-webhook
```
gabemontero, line 125: So this change should remain

pkg/operator/csidriveroperator/csioperatorclient/shared-resource.go
```diff
@@ -9,12 +9,14 @@ const (
 	SharedResourceDriverName             = "csi.sharedresource.openshift.io"
 	envSharedResourceDriverOperatorImage = "SHARED_RESOURCE_DRIVER_OPERATOR_IMAGE"
 	envSharedResourceDriverImage         = "SHARED_RESOURCE_DRIVER_IMAGE"
+	envSharedResourceDriverWebhookImage  = "SHARED_RESOURCE_DRIVER_WEBHOOK_IMAGE"
```
gabemontero, line 12: this change should remain

pkg/operator/csidriveroperator/csioperatorclient/shared-resource.go
```diff
@@ -30,6 +32,8 @@ func GetSharedResourceCSIOperatorConfig() CSIOperatorConfig {
 			"csidriveroperators/shared-resource/07_role_config.yaml",
 			"csidriveroperators/shared-resource/08_rolebinding_config.yaml",
 			"csidriveroperators/shared-resource/11_metrics_service.yaml",
+			"csidriveroperators/shared-resource/13_webhook_service.yaml",
```
gabemontero, line 35: remove these 2

assets/csidriveroperators/shared-resource/09_deployment.yaml
gabemontero: so yeah additionally besides the separate commit remove this file from this PR and include it in a shared resource operator PR per discussion in slack

assets/csidriveroperators/shared-resource/13_webhook_service.yaml
```diff
@@ -0,0 +1,19 @@
+kind: Service
```
gabemontero, line 1: moved to shared resoruce operator PR

assets/csidriveroperators/shared-resource/14_webhook_configmap.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
gabemontero, line 1: move to shared resources operator PR

gabemontero:
OK @jkhelil I have also commented on which changes I think should remain in this PR and which should move to a shared resoruces operator PR

gabemontero:
/lgtm

I checked out the must gather for the e2e-aws-upgrade run and CSO looked fine to me.  Upgrade failed elsewhere.

/assign @jsafrane 

for approval

jkhelil:
/assign @jsafrane 
Can you have a look please

manifests/image-references
```diff
@@ -122,6 +122,10 @@ spec:
...
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-csi-driver-shared-resource-webhook:latest
```
jsafrane, line 128: You must use this name string in CSO's deployment as SHARED_RESOURCE_DRIVER_WEBHOOK_IMAGE env. var, around here:
https://github.com/openshift/cluster-storage-operator/blob/f356bff281fa449ba8bdda35b68e198c2fcfcf92/manifests/10_deployment.yaml#L114-L115

pkg/operator/csidriveroperator/csioperatorclient/shared-resource.go
```diff
@@ -9,12 +9,14 @@ const (
...
 	envSharedResourceDriverOperatorImage = "SHARED_RESOURCE_DRIVER_OPERATOR_IMAGE"
 	envSharedResourceDriverImage         = "SHARED_RESOURCE_DRIVER_IMAGE"
+	envSharedResourceDriverWebhookImage  = "SHARED_RESOURCE_DRIVER_WEBHOOK_IMAGE"
 )
 
 func GetSharedResourceCSIOperatorConfig() CSIOperatorConfig {
 	pairs := []string{
 		"${OPERATOR_IMAGE}", os.Getenv(envSharedResourceDriverOperatorImage),
 		"${DRIVER_IMAGE}", os.Getenv(envSharedResourceDriverImage),
+		"${WEBHOOK_IMAGE}", os.Getenv(envSharedResourceDriverWebhookImage),
```
jsafrane, line 19: I don't see `${WEBHOOK_IMAGE}` string used anywhere in https://github.com/openshift/cluster-storage-operator/blob/master/assets/csidriveroperators/shared-resource/09_deployment.yaml

jsafrane: See how `DRIVER_IMAGE` env. var value is populated in the same file, you must do the same with the webhook image.

gabemontero:
thanks @jsafrane 

yep I missed those spots ^^  in our earlier review @jkhelil

pkg/operator/csidriveroperator/csioperatorclient/shared-resource.go
jkhelil: @jsafrane so here we are bit different than the DRIVER_IMAGE. let me explain
The cluster storage operator is installing the shared resource driver operator, and hence we declare the deployment of the sro into the assets of the cso 
The webhook in an other hand, is installed via the shared resource operator, and its deployment is not known by the cluster storage operator. so Yes I am adding the ${WEBHOOK_IMAGE} string in order to propagate the image of the webhook referenced in the cluster storage operator to the shared resource operator https://github.com/openshift/csi-driver-shared-resource-operator/pull/45/files#diff-f7f6af5f4443f2edda1e6b8ed34f9ac83e4e087eb4df60b345c1437bd2e9466eR20

I am,  may be mistaken the way to propagate the env var, so please correct me if I am wrong

gabemontero: @jkhelil I'm not sure I see what you are getting at, but even though the webhook will be a deployment instead of a daemonset like the driver, the separation of concern wrt this between the cluster storage operator, shared resource operator, and the shared resource repo should be following the same pattern

your PR 45 should have in its deployment something similar to https://github.com/openshift/csi-driver-shared-resource-operator/blob/master/assets/node.yaml#L53 for the image, where it leverages the env var we are propagating from CSO to SRO etc.

gabemontero: and that is what I see at the https://github.com/openshift/csi-driver-shared-resource-operator/pull/45/files#diff-f7f6af5f4443f2edda1e6b8ed34f9ac83e4e087eb4df60b345c1437bd2e9466eR20 item you referenced

I believe @jsafrane corrections are what will make that flow work

gabemontero:
this change in many was is just a subset, copy the pattern, of what @coreydaley did in https://github.com/openshift/cluster-storage-operator/pull/198

pkg/operator/csidriveroperator/csioperatorclient/shared-resource.go
jkhelil: @gabemontero Thank you for the clarification, I think I am getting distured with the cascade of opeartors, but yes you right no difference between DRIVER and WEBHOOK
and here is what is missing @jsafrane : Adding the WEBHOOK_IMAGE string as env var of the shared resource driver operator

gabemontero:
cluster storage operator appears to no be a "no feature freeze" repo @coreydaley @jkhelil 

with our being past feature freeze y'all will need to create a bugzilla and associate it with this PR to drive this home

coreydaley:
/retitle Bug 2086231: BUILD-405: Install the Shared Resource CSI Driver WebHook

PR #279, opened 2022-05-09
Bug 2077599: Alert on vCenter < 7.0.2

jsafrane:
Updated existing ESXi host check to emit info alert when running with a host older that 7.0.2 + added the same alert for vCenter version too.
cc @openshift/storage

### DIFF
--- a/assets/vsphere_problem_detector/12_prometheusrules.yaml
+++ b/assets/vsphere_problem_detector/12_prometheusrules.yaml
@@ -70,12 +70,25 @@ spec:
         # Using min_over_time to make sure the metric is `1` for whole 5 minutes.
         # A missed scraping (e.g. due to a pod restart) will result in prometheus re-evaluating the the alerting rule.
         expr: |
-          min_over_time(vsphere_esxi_version_total{api_version!~"6.7.3|7.*"}[5m]) > 0
+          min_over_time(vsphere_esxi_version_total{api_version=~"^7\\.0\\.[0-1]|^6"}[5m]) > 0
         for: 10m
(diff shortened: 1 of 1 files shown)


jsafrane:
/hold
This may need to wait for OCP 4.12, https://issues.redhat.com/browse/STOR-794

assets/vsphere_problem_detector/12_prometheusrules.yaml
```diff
@@ -70,12 +70,25 @@ spec:
         # Using min_over_time to make sure the metric is `1` for whole 5 minutes.
         # A missed scraping (e.g. due to a pod restart) will result in prometheus re-evaluating the the alerting rule.
         expr: |
-          min_over_time(vsphere_esxi_version_total{api_version!~"6.7.3|7.*"}[5m]) > 0
+          min_over_time(vsphere_esxi_version_total{api_version=~"(^7\\.0\\.[0-1]|^6)"}[5m]) > 0
```
gnufied, line 73: Did you mean this to be - "(^7\\.0\\.[0-1])|^6" ? i.e have "^6" outside the parenthesis

jsafrane: Now that you pointed it out, I think the parenthesis are not needed at all.

jsafrane: Updated a bit, PTAL.

jsafrane:
/hold cancel
We want info alert in 4.11 and raise the severity in 4.12

PR #285, opened 2022-05-16
Revert "Install shared resource csi driver webhook alongside the shared resource csi driver"

dgoodwin:
This reverts commit 0c2757fb0652f78fa5d4c0a4fe343c659404db8c.

The commit in question has broken payload builds with the following
error:

error: failed to push image registry.ci.openshift.org/ocp/release:4.11.0-0.nightly-2022-05-16-100212: unable to upload new layer (0): Patch "https://registry.ci.openshift.org/v2/ocp/release/blobs/uploads/2b297c25-20a4-4811-b97c-bab3bafbf49d?_state=yap-S-67RRqpo0ZcQWfHZz84OreAAOIooCrJFbuNqod7Ik5hbWUiOiJvY3AvcmVsZWFzZSIsIlVVSUQiOiIyYjI5N2MyNS0yMGE0LTQ4MTEtYjk3Yy1iYWIzYmFmYmY0OWQiLCJPZmZzZXQiOjAsIlN0YXJ0ZWRBdCI6IjIwMjItMDUtMTZUMTA6MTA6NTAuOTc5NjQxNjkyWiJ9": operator "cluster-storage-operator" contained an invalid image-references file: no input image tag named "csi-driver-shared-resource-webhook"

See: https://amd64.ocp.releases.ci.openshift.org/releasestream/4.11.0-0.nightly/release/4.11.0-0.nightly-2022-05-16-100212

Per TRT policy we are asked to revert breaking changes as quickly as possible to get payloads flowing again, apologies for any inconvenience. We ask that to re-introduce, you revert *this* PR, and layer an additional separate commit on top that addresses the problem. However in this case that may not be applicable as I suspect you need to fix something in the ART pipelines to get your new image tag, then reintroduce this change as is?

Will file a bug for tracking purposes to make sure the change doesn't get lost.

### DIFF
--- a/assets/csidriveroperators/shared-resource/09_deployment.yaml
+++ b/assets/csidriveroperators/shared-resource/09_deployment.yaml
@@ -25,8 +25,6 @@ spec:
           value: ${DRIVER_IMAGE}
         - name: NODE_DRIVER_REGISTRAR_IMAGE
           value: ${NODE_DRIVER_REGISTRAR_IMAGE}
-        - name: WEBHOOK_IMAGE
-          value: ${WEBHOOK_IMAGE}
         - name: POD_NAME
           valueFrom:
             fieldRef:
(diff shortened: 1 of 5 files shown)


dgoodwin:
Normally our policy is revert no matter what, but in this case because there's no fix that can be added to the PR, we really just need to try again, I see no point in pushing for the revert, so yes we can pause in this case.

/hold

jkhelil:
> Normally our policy is revert no matter what, but in this case because there's no fix that can be added to the PR, we really just need to try again, I see no point in pushing for the revert, so yes we can pause in this case.
> 
> /hold

thank you @dgoodwin

yselkowitz:
We have 4.11 nightly builds again.
/close

PR #290, opened 2022-06-20
Bug 2077599: Fix vCenter / ESXi version alerts

jsafrane:
The alerting regular expression must match the whole label from `^` to `$`, hence add `.*` at the end.


jsafrane:
cc @openshift/storage

PR #292, opened 2022-06-28
Bug 2101645: DefaultStorageClassController reports fake message on azure and openstack

RomanBednar:
Letting newStorageClassForCluster() default to unsupportedPlatformError is not appropriate because later in sync this results in reporting `"No default StorageClass for this platform"`. Instead we should report this message for Azure and OpenStack:

 `"StorageClass provided by supplied CSI Driver instead of the cluster-storage-operator"`

### DIFF
--- a/pkg/operator/defaultstorageclass/controller.go
+++ b/pkg/operator/defaultstorageclass/controller.go
@@ -196,6 +196,10 @@ func newStorageClassForCluster(infrastructure *configv1.Infrastructure) (*storag
 		storageClassFile = "storageclasses/gcp.yaml"
 	case configv1.VSpherePlatformType:
 		storageClassFile = "storageclasses/vsphere.yaml"
+	case configv1.AzurePlatformType:
+		return nil, supportedByCSIError
+	case configv1.OpenStackPlatformType:
+		return nil, supportedByCSIError
 	case configv1.OvirtPlatformType:
 		return nil, supportedByCSIError
 	default:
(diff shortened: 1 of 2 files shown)


pkg/operator/defaultstorageclass/controller.go
```diff
@@ -196,6 +196,10 @@ func newStorageClassForCluster(infrastructure *configv1.Infrastructure) (*storag
 		storageClassFile = "storageclasses/gcp.yaml"
 	case configv1.VSpherePlatformType:
 		storageClassFile = "storageclasses/vsphere.yaml"
+	case configv1.AzurePlatformType:
```
Phaow, line 199: Seems the unit test also need to adapt the change.
Need revert: https://github.com/openshift/cluster-storage-operator/commit/0d93ba2a28faafe0f4cd587c3e48f45a83cab961

RomanBednar: Indeed, reverted.

pkg/operator/defaultstorageclass/controller.go
```diff
@@ -196,6 +196,10 @@ func newStorageClassForCluster(infrastructure *configv1.Infrastructure) (*storag
...
+		return nil, supportedByCSIError
```
jsafrane, line 200: Hold on, should it create in-tree StorageClass for Azure File? Its CSI migration is not GA yet.

jsafrane: Scratch this, we did not create StorageClass for Azure File

PR #299, opened 2022-07-11
Bug 2098655: gcp cluster rollback fails due to storage failure

RomanBednar:
Bump library-go so that we include [this](https://github.com/openshift/library-go/pull/1355) change from library-go in gcp driver operator. This should allow changing/re-creating the StorageClass allowing downgrade from 4.11 to 4.10 which now fails because the re-creation is not allowed.

cc @openshift/storage


jsafrane:
/lgtm
/approve
But you need 4.12 BZ for this.

PR #300, opened 2022-07-14
Bug 2098655: gcp cluster rollback fails due to storage failure

RomanBednar:
Bump library-go so that we include [this](https://github.com/openshift/library-go/pull/1355) change from library-go in gcp driver operator. This should allow changing/re-creating the StorageClass allowing downgrade from 4.11 to 4.10 which now fails because the re-creation is not allowed.

cc @openshift/storage

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/openshift/api v0.0.0-20220114150019-2499da51153e
 	github.com/openshift/build-machinery-go v0.0.0-20211213093930-7e33a7eb4ce3
 	github.com/openshift/client-go v0.0.0-20211209144617-7385dd6338e3
-	github.com/openshift/library-go v0.0.0-20220114151217-4362aa519714
+	github.com/openshift/library-go v0.0.0-20220630204433-c71d40c7de49
 	github.com/prometheus-operator/prometheus-operator v0.44.1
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -85,27 +85,27 @@ require (
...
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
-	k8s.io/apiserver v0.23.0 // indirect
-	k8s.io/kube-aggregator v0.23.0 // indirect
-	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65 // indirect
-	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.25 // indirect
+	k8s.io/apiserver v0.24.0 // indirect
+	k8s.io/kube-aggregator v0.24.0 // indirect
```
jsafrane, line 104: This looks wrong, 4.10 should use 0.23. Did you bump library-go from its release-4.10 branch?

RomanBednar: Probably not, now it should be ok.

Phaow:
/label cherry-pick-approved
The failed vsphere test snapshot is not support on 4.10.

PR #303, opened 2022-07-28
[CNF-5645] add storage capability annotation

qJkee:
add annotation, so operator can be configured as optional


gnufied:
Btw storage team was wondering if hypershift environments will ever require disabling storage or snapshot features. If that is the case hypershift design must consider these capabilities so as they can be optionally enabled or disabled.

qJkee:
@gnufied 
hypershift team replied that they don't see any issues with that feature from their side
/cc @jsafrane

bertinatto:
Link for the enhancement (for folks what's this about): https://github.com/openshift/enhancements/blob/master/enhancements/installer/component-selection.md

This looks useful, I wish we had this earlier on.

I'll let other tag it, but this PR LGTM.

manifests/06_operator_cr.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   annotations:
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
-    include.release.openshift.io/single-node-developer: "true"
+    capability.openshift.io/name: Storage
```
jsafrane, line 8: By removing `include.release.openshift.io/single-node-developer: "true"`, the storage operator will **never** be installed in SNO cluster in any cloud and thus probably break all StatefulSet tests in CI.

IMO, we should keep `include.release.openshift.io/single-node-developer: "true"` there, Storage will be installed by default, however, customers will have option to remove it if they need to trim down the cluster.

qJkee: hmm, I saw an example PR where it was removed but now can't find it.
I'll return this lines as well then

qJkee: Fixed. Do we need capability annotations for IBM-related resources? Are they still managed by CVO?

duanwei33:
Pre-merge verification for https://issues.redhat.com/browse/CNF-5645. (API and installer PR are already merged)

- a) with pre-merge PR, building cluster with **_baselineCapabilitySet: None_** in install-config:
```
    capabilities:
      baselineCapabilitySet: None
```

The Storage CO is not installed:
```
$ oc get co storage
Error from server (NotFound): clusteroperators.config.openshift.io "storage" not found
```


- b) with pre-merge PR, building cluster with **_baselineCapabilitySet: None_** and **_enabledCapabilities: ["Storage"]_** in install-config:
```
    capabilities:
      additionalEnabledCapabilities:
      - Storage
      baselineCapabilitySet: None
```

The Storage CO is present:
```
$ oc get co storage
NAME      VERSION                                                   AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
storage   4.12.0-0.ci.test-2022-08-30-012524-ci-ln-4lntyc2-latest   True        False         False      37m     

$ oc get clusterversion version -ojson | jq .status.capabilities
{
  "enabledCapabilities": [
    "Storage"
  ],
  "knownCapabilities": [
    "Console",
    "Insights",
    "Storage",
    "baremetal",
    "marketplace",
    "openshift-samples"
  ]
}
```

Also, performing regression test for case b) which looks ok. 

So looks good from QE side.

duanwei33:
The qe-approved added, but if there is anything else need to be tested, please let me know.

manifests/02_csi_driver_operators_namespace.yaml
```diff
@@ -1,3 +1,4 @@
+# We keep this namespace without capability name untill we merge cluster-csi-snapshot-controller capability
```
jsafrane, line 1: Namespace `openshift-cluster-csi-drivers` is not used by snapshot-controller, so this comment is wrong.

Question is, if CVO should create this namespace without `Storage` capability. It's used by optional CSI drivers, like AWS EFS or GCP Filestore, installed via OLM. IMO OLM should be able to create the namespace and therefore this namespace should have `Storage` capability.

@openshift/storage what do you think?

manifests/06_operator_cr.yaml
jsafrane: Yes, we still need annotations for IBM-related resources.

qJkee: got it, i'll keep them

manifests/02_csi_driver_operators_namespace.yaml
qJkee: If we go this way I would need to provide a namespace creation mechanism/rule to OLM, right?

jsafrane: OLM already knows how to create a namespace.

qJkee: fixed

duanwei33:
@qJkee @jsafrane I did another try with the latest commit and double check the resource defined in manifest file, others are good but it seems we need to add the capability for [03_credentials_request_powervs.yaml](https://github.com/openshift/cluster-storage-operator/blob/master/manifests/03_credentials_request_powervs.yaml) which is added recently and not in this PR.

duanwei33:
@jsafrane Another question is do we support users installing community CSI Driver? Like should we cover the test for community CSI Driver + our csi-snapshot-controller?

qJkee:
> @qJkee @jsafrane I did another try with the latest commit and double check the resource defined in manifest file, others are good but it seems we need to add the capability for [03_credentials_request_powervs.yaml](https://github.com/openshift/cluster-storage-operator/blob/master/manifests/03_credentials_request_powervs.yaml) which is added recently and not in this PR.

Thank you. I've added it as well

jsafrane:
@duanwei33 sorry, I missed your previous question

>  Another question is do we support users installing community CSI Driver? Like should we cover the test for community CSI Driver + our csi-snapshot-controller?

We do support community CSI drivers, installed e.g. through OLM, + their snapshots. And it would be nice to test it.

duanwei33:
Sure, I will test it in another thread, from my understanding, the PR is okay to merge from QE side.

duanwei33:
@qJkee @jsafrane 
Hi, do you know how we could get it merged? I think we could start automation in our CI if merged.

jsafrane:
It needs docs + PX approval. PX is easy once documentation is in place.

qJkee:
/hold
untill i understand the connection with snapshot controller

jerpeter1:
We have planned docs for this, so adding docs-approved, as per the response we got from the docs team

/label docs-approved

PR #305, opened 2022-08-02
Bug 2108054: Add alert about unsupported CSI driver

jsafrane:
Add info alert that upstream vSphere CSI driver is installed.

cc @openshift/storage


jsafrane:
/hold
I need to fix library-go first to handle PrometheusRules better.

jsafrane:
library-go updated
/hold cancel

bertinatto (commented):
\\lgtm

PR #307, opened 2022-08-08
STOR-956: Bump(openshift/api): to get CSI changes

bertinatto:
CC @openshift/storage


bertinatto:
/label px-approved
/labe docs-approved
This should simply allow the creation of a ClusterCSIDriver CR with a new name for GCP Filestore.

PR #308, opened 2022-08-17
Bug 2108054: Allow Prometheus to scan openshift-cluster-csi-drivers namespace

jsafrane:
vmware-vsphere-csi-driver-operator emits metrics, so CSO must give Prometheus permissions to scan its namespace.

cc @openshift/storage


jsafrane:
In https://github.com/openshift/cluster-storage-operator/pull/309 I tested that there is `UnsupportedCSIDriverInstalled` alert firing when the cluster is *installed* with CSIDriver instance without OCP annotation (which simulates upstream CSI driver installation).

bertinatto (commented):
/lgtm

PR #318, opened 2022-09-27
STOR-1039: Make changes into CSO for hypershift

gnufied:
Create new driver starter for hypershift
Add code to create hypershift clusters


jsafrane:
`
"csidriveroperators/aws-ebs/mgmt_assets/01_operator_role.yaml\" (string): Role.rbac.authorization.k8s.io \"aws-ebs-csi-driver-operator-role\" is invalid: rules[5].verbs: Required value: verbs must contain at least one value
`

pkg/operator/csidriveroperator/hypershift_deployment_controller.go
```diff
@@ -0,0 +1,183 @@
...
+	factory           *factory.Factory
+}
+
+var _ factory.Controller = &HyperShiftDeploymentController{}
+
+var (
+	envHyperShiftImage = os.Getenv("HYPERSHIFT_IMAGE")
+)
+
+func NewHyperShiftControllerDeployment(
```
jsafrane, line 56: If we remove `if !olmRemovalComplete {...}` from CSIDriverOperatorDeploymentController, then both HyperShiftControllerDeployment + CSIDriverOperatorDeploymentController are just library-go DeploymentControllers with extra  hooks.

IMO, we can safely remove OLMOperatorRemovalController now. and use DeploymentController with hooks. The code will get much simpler - *DriverStarter will just start DeploymentControllers with the right hooks.

pkg/operator/csidriveroperator/driverstarter_hypershift.go
```diff
@@ -0,0 +1,231 @@
...
+			})
+			klog.V(2).Infof("Starting ControllerManager for %s", ctrl.operatorConfig.ConditionPrefix)
+			go ctrl.mgr.Start(ctx)
+			ctrl.running = true
+		}
+	}
+	return nil
+}
+
+func (c *CSIDriverStarterControllerHyperShift) createCSIControllerManager(
```
jsafrane, line 170: Does it make sense to have two *DriverStarters instead of one? Their code before starting a CSI driver operator looks quite same, the DriverStarter could have just different functions `createStandaloneCSIControllerManager` / `createHyperShiftCSIControllerManager`.

I don't mind running two StaticResourceControllers in the standalone cluster.

assets/csidriveroperators/aws-ebs/guest_assets/10_cr.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: operator.openshift.io/v1
```
jsafrane, line 1: nits:

* remove `assets` from `guest_assets` / `mgmt_asset` - they're in `assets` dir already.
* directory names should have `hypershift` somewhere, `assets/csidriveroperators/aws-ebs/hypershift/guest`?

assets/csidriveroperators/aws-ebs/mgmt_assets/01_operator_role.yaml
```diff
@@ -0,0 +1,86 @@
+# RBAC needed to start the AWS Operator in mgmt cluster
```
jsafrane, line 1: I like separate assets dirs, we could have multiple copies of assets for now, but in the long run we should have a way how to generate hypershift assets from standalone or the other way around.

pkg/operator/standalone_starter.go
```diff
@@ -0,0 +1,119 @@
...
+		controllerConfig.EventRecorder,
+	)
+
+	countStorageClasses(storageClassController, clients)
+
+	// We dont' need this in hypershift guest or mgmt clusters
+	snapshotCRDController := snapshotcrd.NewController(
+		clients,
+		controllerConfig.EventRecorder,
+	)
```
jsafrane, line 47: Can we remove the whole snapshotCRDController now?

IMO, CSISnapshotControllerOperator will get Degraded if it can't apply VolumeSnapshot CRDs. And it will overwrite any v1alpha1 CRDs.

pkg/operator/csidriveroperator/hypershift_deployment_controller.go
```diff
@@ -0,0 +1,183 @@
...
+	targetVersion     string
+	eventRecorder     events.Recorder
+	infraLister       configv1listers.InfrastructureLister
+	factory           *factory.Factory
+}
+
+var _ factory.Controller = &HyperShiftDeploymentController{}
+
+var (
+	envHyperShiftImage = os.Getenv("HYPERSHIFT_IMAGE")
```
jsafrane, line 53: This env. var should be set in also in `manifests/10_deployment-hypershift.yaml`, so we can copy the yaml file to HyperShift.

assets/csidriveroperators/aws-ebs/mgmt_assets/01_operator_role.yaml
```diff
@@ -0,0 +1,77 @@
...
+  - configmaps
+  - secrets
+  verbs:
+  - '*'
+- apiGroups:
+  - ''
+  resources:
+  - namespaces
+  verbs:
+  - get
```
jsafrane, line 24: This is IMO not needeed any longer.

gnufied: removed.

assets/csidriveroperators/aws-ebs/mgmt_assets/01_operator_role.yaml
gnufied: Yeah I was thinking that - we could use `kustomize` or another tool to DRY up the YAMLs, so as we don't have to make duplicate changes all over. But I would prefer to avoid using golang to mutate these YAMLs unless we really need some runtime data.

jsafrane:
/label docs-approved
In theory, this is not an user facing change. HyperShift should work the same as before, just some components (CSO, AWS EBS CSI driver operator, control plane of the CSI driver) will move from guest to mgmt cluster during upgrade without any effect on functionality.

pkg/operator/csidriveroperator/driverstarter_hypershift.go
gnufied: There are still enough difference between them that, I would like to keep them separate for now until we can think of a cleaner way of reducing duplication in the code. I sort of don't like superfluous conditions being added to a list which is already pretty long.  I sort of prefer some duplication until we have cleaner design. This keeps standalone cluster code almost untouched.

pkg/operator/csidriveroperator/hypershift_deployment_controller.go
gnufied: Fixed.

PR #319, opened 2022-09-27
STOR-1050: bump openshift/api to latest master

gnufied:
Bump api to get latest CRD.


bertinatto (commented):
/lgtm

duanwei33:
checked with the following PRs:
launch openshift/vmware-vsphere-csi-driver-operator#107,openshift/cluster-version-operator#835,openshift/cluster-storage-operator#319 vsphere

```
$ oc explain clustercsidrivers.spec.driverConfig
KIND:     ClusterCSIDriver
VERSION:  operator.openshift.io/v1

RESOURCE: driverConfig <Object>

DESCRIPTION:
     driverConfig can be used to specify platform specific driver configuration.
     When omitted, this means no opinion and the platform is left to choose
     reasonable defaults. These defaults are subject to change over time.

FIELDS:
   driverType	<string> -required-
     driverType indicates type of CSI driver for which the driverConfig is being
     applied to. Valid values are:
     * vSphere Allows configuration of vsphere CSI driver topology. ---
     Consumers should treat unknown values as a NO-OP.

   vSphere	<Object>
     vsphere is used to configure the vsphere CSI driver.
```

PR #324, opened 2022-10-18
Bug 2093016: Add alert about attach / mount failing

jsafrane:
Show an alert when all attach or mount operations for a volume plugin (or CSI driver) are failing on a node for 5 minutes. A single success will make the alert go away.

cc @openshift/storage

### DIFF
--- a/manifests/12_prometheusrules.yaml
+++ b/manifests/12_prometheusrules.yaml
@@ -25,3 +25,22 @@ spec:
             Cluster storage operator monitors all storage classes configured in the cluster
             and checks there is not more than one default StorageClass configured.
           message: "StorageClass count check is failing (there should not be more than one default StorageClass)"
+
+    - name: storage-operations.rules
+      rules:
+      - alert: PodStartupStorageOperationsFailing
+        # There was at least one failing operation in past 5 minutes *and* there was no successful one.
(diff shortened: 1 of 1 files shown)


manifests/12_prometheusrules.yaml
```diff
@@ -25,3 +25,22 @@ spec:
             Cluster storage operator monitors all storage classes configured in the cluster
             and checks there is not more than one default StorageClass configured.
           message: "StorageClass count check is failing (there should not be more than one default StorageClass)"
+
+    - name: storage-operations.rules
+      rules:
+      - alert: PodStartupStorageOperationsFailing
+        # There was at least one failing operation in past 5 minutes *and* there was no successful one.
+        # Focus on attach and mount operations - they have the same diagnostic steps and are the most common.
+        expr: |
```
gnufied, line 34: We should perhaps add a comment that - this does not cover the cases where attach operation does not even start. For example - a node is shutdown and pod is deleted, but detach won't start until a certain check in ADC expires (I think something like 5 minutes after which volume is force detached).

Which reminds me - do we need an alert for detach failures? may be we can do that in a follow up. But if detach is failing then attach operation may not even start and hence no attach failure metric might be emitted.

PR #325, opened 2022-10-26
OCPBUGS-2880: Fix panic on empty featureGate.spec.customNoUpgrade

jsafrane:
Don't panic when user sets `fg.Spec.FeatureSet`, but leaves `fg.Spec.CustomNoUpgrade` nil or empty. It is a weird configuration, where user wants custom flags but does not set any, but the operator should not crash.

cc @openshift/storage


bertinatto (commented):
/lgtm

PR #326, opened 2022-11-04
Add infra to support running unit tests against prom rules

gnufied:


alert_rule_tests/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+    https://kislyuk.github.io/yq/
+
+## Extract rules from k8s asset:
+    cat assets/vsphere_problem_detector/12_prometheusrules.yaml|yq -Y '.spec' > alert_rule_tests/12_prometheusrules.yaml
+
+## Run the unit test:
+
+
+   promtool test rules ./alert_rule_tests/vsphere_older_version_test.yaml
```
jsafrane, line 13: This should be part of `make test` or `make test-alerts` and called in CI.

alert_rule_tests/README.md
```diff
@@ -0,0 +1,13 @@
+# How to test and run prometheus rules
+
+## Download and install yq tool
+
+    https://kislyuk.github.io/yq/
+
+## Extract rules from k8s asset:
+    cat assets/vsphere_problem_detector/12_prometheusrules.yaml|yq -Y '.spec' > alert_rule_tests/12_prometheusrules.yaml
```
jsafrane, line 8: I found multiple references to `yq` in openshift/release repo, it may be available during `make test`. Therefore we may not need to store `alert_rule_tests/12_prometheusrules.yaml` in git.

gnufied: Not in the default image at least:

```
cat assets/vsphere_problem_detector/12_prometheusrules.yaml|yq -Y '.spec' > alert_rule_tests/12_prometheusrules.yaml
/bin/sh: yq: command not found
```

I also think we will have to pull ins special image for making `promtool` available..

PR #327, opened 2022-11-07
OCPBUGS-3312: Add capability annotation to openshift-cluster-storage-operator namespace

bertinatto:
The namespace should only be created when both Storage and CSISnapshot capabilities are enabled.

CC @openshift/storage

### DIFF
--- a/manifests/0000_49_cluster-storage-operator_01_operator_namespace.yaml
+++ b/manifests/0000_49_cluster-storage-operator_01_operator_namespace.yaml
@@ -9,6 +9,7 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     openshift.io/node-selector: ""
     workload.openshift.io/allowed: "management"
+    capability.openshift.io/name: Storage+CSISnapshot
   labels:
     openshift.io/cluster-monitoring: "true"
   name: openshift-cluster-storage-operator


jsafrane:
> The namespace should only be created when both Storage and CSISnapshot capabilities are enabled.

The namespace should be created only when at least _one of_ Storage or CSISnapshot is enabled. Can CVO handle such logic?

bertinatto:
> > The namespace should only be created when both Storage and CSISnapshot capabilities are enabled.
> 
> The namespace should be created only when at least _one of_ Storage or CSISnapshot is enabled. Can CVO handle such logic?

I misinterpreted the enhancement. Currently this doesn't seem to be possible.

@bparees, since you're the enhancement's author, can you confirm that?

bparees:
> > The namespace should be created only when at least one of Storage or CSISnapshot is enabled. Can CVO handle such logic?
> I misinterpreted the enhancement. Currently this doesn't seem to be possible. Checked both the [enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/installer/component-selection.md) and the code.
> @bparees, since you're the enhancement's author, can you confirm that?

You're correct, you can define a resource that will only be reconciled when all of a set of caps are enabled.  You cannot define a resource that will be reconciled if "any one of N" caps are enabled.

bertinatto:
/hold
Until we sort out https://github.com/openshift/cluster-storage-operator/pull/327#issuecomment-1307472481.

bertinatto:
/close
xref: https://issues.redhat.com/browse/OCPBUGS-3312

PR #328, opened 2022-11-28
OCPBUGS-4173: Restart vsphere-problem-detector after credentials change

jsafrane:
Replace custom VSphereProblemDetectorDeploymentController with common DeploymentController and add cloud-config ConfigMap and vSphere credentials Secret hashes to the vsphere-problem-detector Deployment.

As result, vsphere-problem-detector is restarted after credentials change and thus should check the new credentials or cloud config quickly (= in 2-3 minutes, because of leader election). Therefore, any alert about bad credentials will be cleared 2-3 minutes after user provides the right ones.

@openshift/storage


pkg/operatorclient/operatorclient.go
```diff
@@ -95,3 +95,52 @@ func (c OperatorClient) SetObjectAnnotations(required map[string]string) error {
 	}
 	return nil
 }
+
+func (c OperatorClient) EnsureFinalizer(ctx context.Context, finalizer string) error {
```
bertinatto, line 99: Any reasons we're not using the generic operator client here?
Regardless, not for this PR.

pkg/operator/vsphereproblemdetector/vsphere_problem_detector_starter.go
```diff
@@ -129,3 +170,64 @@ func (c *VSphereProblemDetectorStarter) createVSphereProblemDetectorManager(
...
+			"${LOG_LEVEL}", strconv.Itoa(logLevel),
+		}
+
+		replacer := strings.NewReplacer(pairs...)
+		newDeployment := replacer.Replace(string(deployment))
+		return []byte(newDeployment), nil
+	}
+}
+
+func (c *VSphereProblemDetectorStarter) WithConfigMapHashAnnotationHook(namespace string, cmInformer coreinformers.ConfigMapInformer) deploymentcontroller.DeploymentHookFunc {
```
bertinatto, line 188: We're probably going to need this in library-go, but we can do this later.

pkg/operator/vsphereproblemdetector/vsphere_problem_detector_starter.go
```diff
@@ -2,31 +2,45 @@ package vsphereproblemdetector
...
+	infraConfigName                     = "cluster"
+	vSphereProblemDetectorOperatorImage = "VSPHERE_PROBLEM_DETECTOR_OPERATOR_IMAGE"
+	secretName                          = "vsphere-cloud-credentials"
+	cloudConfigNamespace                = "openshift-config"
 )
 
 type VSphereProblemDetectorStarter struct {
 	controller     manager.ControllerManager
-	operatorClient *operatorclient.OperatorClient
+	operatorClient v1helpers.OperatorClientWithFinalizers
```
bertinatto, line 43: Why do you need `OperatorClientWithFinalizers`? It doesn't seem like you're using it the extra methods.

jsafrane: `DeploymentController` needs it, because it supports removable operands.

pkg/operatorclient/operatorclient.go
jsafrane: I tried to use generic operator client from https://github.com/openshift/library-go/blob/master/pkg/operator/genericoperatorclient/dynamic_operator_client.go, but it does not support `SetObjectAnnotations()` used in [OLMOperatorRemovalController](https://github.com/openshift/cluster-storage-operator/blob/53afa77f6bbca7cfe28b914930beff2b3c407d93/pkg/operator/csidriveroperator/olmremovalcontroller.go#L262)

Removing `OLMOperatorRemovalController` could be a way forward.

pkg/operator/vsphereproblemdetector/vsphere_problem_detector_starter.go
```diff
@@ -115,12 +129,39 @@ func (c *VSphereProblemDetectorStarter) createVSphereProblemDetectorManager(
...
-		clients,
-		c.versionGetter,
-		c.targetVersion,
+	deploymentAssets, err := assets.ReadFile("vsphere_problem_detector/07_deployment.yaml")
+	if err != nil {
+		panic(err)
+	}
+
+	mgr = mgr.WithController(deploymentcontroller.NewDeploymentController(
+		"SphereProblemDetectorDeploymentController",
```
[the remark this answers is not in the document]
jsafrane: Fixed

bertinatto (commented):
/lgtm

PR #331, opened 2023-01-25
Hypershift changes for PowerVS CSI driver operator

dharaneeshvrd:
Fixes https://issues.redhat.com/browse/MULTIARCH-3203


dharaneeshvrd:
@gnufied @jsafrane Could you please review this PR when you get a chance?

dharaneeshvrd:
@gnufied @jsafrane Could you please review this PR when you get a chance?

gnufied:
Has these changes been tested in a cluster btw?

assets/csidriveroperators/powervs-block/hypershift/mgmt/06_deployment.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: apps/v1
```
gnufied, line 1: There are is already IBM managed cloud YAML in `manifest` directory? When we originally implemented these changes , we generated a hypershift specific YAML - https://github.com/openshift/cluster-storage-operator/blob/master/manifests/10_deployment-hypershift.yaml

Is the code that will initialize CSO is going to be in control-plane-operator? Are we planning to sync those assets with YAMLs in `manifest` ?

dharaneeshvrd: Yes these changes are trying to address the change in CSO being running in control plane.
Currently if CSO installed in control plane, nothing happening for PowerVS platform, since it does not handled [here](https://github.com/openshift/cluster-storage-operator/blob/master/pkg/operator/starter.go#L68). With this changes we have tried to addressed that.

Instantiating CSO will be taken care by hypershift control plane operator itself, there is no change required in how CSO is getting installed depending on the platform. Only few changes were required for CSI driver operator and CSI driver since its moved control plane (mounting config and creds).  

We have tested this flow with changes made in CSO, PowerVS CSI operator and CSI driver and its working fine.

assets/csidriveroperators/powervs-block/hypershift/mgmt/06_deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+kind: Deployment
+metadata:
+  name: powervs-block-csi-driver-operator
+  namespace: ${CONTROLPLANE_NAMESPACE}
+  annotations:
+    config.openshift.io/inject-proxy: powervs-block-csi-driver-operator
```
gnufied, line 7: Can you check if we need this proxy? https://github.com/openshift/cluster-storage-operator/pull/337/files

dharaneeshvrd: We don't need this proxy, tested without it and it worked fine. 
So removed it.

mkumatag:
/cc @juliemathew 

for QE ack.

PR #332, opened 2023-01-26
STOR-1128: Remove defaultstorageclass controller

bertinatto:
Since CSI migration is GA for both vSphere and Azure (OCP 4.13), CSO does not need to create any StorageClasses anymore. As a result, we no longer need this controller.

CC @openshift/storage 

/hold
for https://github.com/openshift/kubernetes/pull/1432


bertinatto:
I tested this with https://github.com/openshift/cluster-storage-operator/pull/134 and the default StorageClass was:

1.  **thin** when upgraded from 4.12
2. **thin-csi** on a new cluster

dobsonj:
@bertinatto this PR depends on https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/134 doesn't it?
e2e-vsphere-csi is failing because prometheus depends on the default SC, and there isn't one without that PR.

bertinatto:
> @bertinatto this PR depends on [openshift/vmware-vsphere-csi-driver-operator#134](https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/134) doesn't it? e2e-vsphere-csi is failing because prometheus depends on the default SC, and there isn't one without that PR.

Right, see https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/134#issuecomment-1406584402. QE is currently verifying both PRs together.

Putting this on hold to avoid confusion:

/hold

PR #334, opened 2023-02-02
STOR-1123: Add alert about topology misconfiguration

jsafrane:
Show alert that topology is configured both in Infrastructure and ClusterCSIDriver.

cc @openshift/storage

### DIFF
--- a/assets/csidriveroperators/vsphere/15_prometheusrules.yaml
+++ b/assets/csidriveroperators/vsphere/15_prometheusrules.yaml
@@ -28,3 +28,20 @@ spec:
             including its Deployment, DaemonSet and CSIDriver objects and OpenShift will
             automatically install a supported version of the CSI driver.
           message: "An unsupported version of vSphere CSI driver installation detected."
+      - alert: VSphereTopologyMisconfiguration
+        # min() takes minimum over both source="infrastructure" and source="clustercsidriver"
+        # dimensions and only when both are above zero the alert is triggered.
+        expr:  min(vsphere_topology_tags{source=~"infrastructure|clustercsidriver"}) > 0
+        for: 10m
+        labels:
+          severity: info
+        annotations:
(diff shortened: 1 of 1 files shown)


jsafrane:
/hold
for https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/135 to be merged first

gnufied:
Did you manage to test the alert btw?

jsafrane:
Yes, I did

assets/csidriveroperators/vsphere/15_prometheusrules.yaml
```diff
@@ -28,3 +28,19 @@ spec:
             including its Deployment, DaemonSet and CSIDriver objects and OpenShift will
             automatically install a supported version of the CSI driver.
           message: "An unsupported version of vSphere CSI driver installation detected."
+      - alert: VSphereTopologyMisconfiguration
+        # Both source="infrastructure" and source="clustercsidriver" must be above 0
+        expr:  min(vsphere_topology_tags{source=~"infrastructure|clustercsidriver"}) > 0
```
gnufied, line 33: I feel like a query ex-plainer is in order, because it is a clever query!

> min() works by taking minimum over both `infrastructure` and `clustercsidriver` dimensions and only when both are 1, then value will be greater than  0 and an alert will be triggered.

I think  we should add this comment.

jsafrane: rephrased your comment a bit, PTAL

duanwei33:
see verify step in https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/135#issuecomment-1424220394

jsafrane:
/label px-approved
/label docs-approved
We do not document all our alerts in docs, this is only Info one.

PR #340, opened 2023-02-15
STOR-1128: Stop creating an in-tree SC for vSphere

bertinatto:
This PR reworks #332, which was reverted because CSO currently expects at least one controller that sets Available & Progressing conditions.

CC @openshift/storage


gnufied:
The failures look unrelated.

/lgtm

PR #344, opened 2023-02-21
OCPBUGS-6992: Add UID to CSO Pod to be able to run with custom SCCs

jsafrane:
When there is a custom SCC with `runAsUser: type: runAsNonRoot`, CSO pod can match it. Then the Pod must have a non-root UID assigned to be able to run.

This moves the CSO Pods from `restricted-v2` SCC to `nonroot-v2` SCC in the default OCP installation.

This is a workaround for CSO using cluster-admin role, which can allow CSO to use any SCC. If CSO had specific role(s) to use only few selected SCCs, we could remove this patch.

### DIFF
--- a/manifests/10_deployment-hypershift.yaml
+++ b/manifests/10_deployment-hypershift.yaml
@@ -125,6 +125,7 @@ spec:
           name: guest-kubeconfig
       securityContext:
         runAsNonRoot: true
+        runAsUser: 11411
         seccompProfile:
           type: RuntimeDefault
       serviceAccountName: cluster-storage-operator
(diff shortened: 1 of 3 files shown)


jsafrane:
cc @openshift/storage

bertinatto (commented):
/lgtm

jmcmeek:
@jsafrane I've seen the same behavior with the csi-snapshot-controller-operator pods after an upgrade from 4.11 to 4.12.  I was going to ask if you knew if it had been changed, but I think I found the source - https://github.com/openshift/cluster-csi-snapshot-controller-operator/tree/master/manifests

I'm working with an IBM Cloud OpenShift customer, but it looks like all the `07_deployment*.yaml` versions should behave the same, and all could result in pods in CreateContainerConfigError due to using an unexpected SCC.

I don't have access to the jira issue.  Do you know you know if the csi-snapshot-controller-operator is going to be changed as well?

jsafrane:
/cherry-pick release-4.12

@jmcmeek yes, I will fix snapshot opretat in 4.12 too

PR #352, opened 2023-03-20
OCPBUGS-8691: Hypershift: set control plane operand properties

jsafrane:
Set affinity, tolerations, node selector and priority rules for operands in the hosted control plane.


bertinatto (commented):
/lgtm

PR #358, opened 2023-04-11
OCPBUGS-11617: Bump github.com/prometheus-operator/prometheus-operator

jsafrane:
To fix CVE-2021-36157 - new prometheus-operator does not import grafana / cortex at all.

This bumps also Kubernetes libs as a side-effect.

Also bump go.opentelemetry.io libs to fix dependency issues.

@openshift/storage


jsafrane:
This fixes also OCPBUGS-12593

PR #362, opened 2023-04-19
WIP: CCO-324: azure-disk: mount OIDC token and specify serviceAccountName

jstuever:
Mount the OIDC token into the `azure-disk-csi-driver-operator` and specify the `serviceAccountName` in the `azure-disk-csi-driver-operator` credentials request.

https://issues.redhat.com/browse/CCO-324

### DIFF
--- a/assets/csidriveroperators/azure-disk/08_deployment.yaml
+++ b/assets/csidriveroperators/azure-disk/08_deployment.yaml
@@ -52,6 +52,10 @@ spec:
           requests:
             memory: 50Mi
             cpu: 10m
+        volumeMounts:
+        - name: bound-sa-token
+          mountPath: /var/run/secrets/openshift/serviceaccount
+          readOnly: true
       priorityClassName: system-cluster-critical
       serviceAccountName: azure-disk-csi-driver-operator
       nodeSelector:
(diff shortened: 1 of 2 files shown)


manifests/03_credentials_request_azure.yaml
```diff
@@ -8,6 +8,8 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     capability.openshift.io/name: Storage
 spec:
+  serviceAccountNames:
+  - azure-disk-csi-driver-operator
```
RomanBednar, line 12: @jstuever If we need to use the secret with [this](https://github.com/openshift/azure-disk-csi-driver-operator/blob/f71f1449938a1dde07904dc079a3f36284f8e99a/assets/controller.yaml#L26) service account instead, does it need to be added here?

We have to load the secret with [injector](https://github.com/openshift/azure-disk-csi-driver-operator/blob/f71f1449938a1dde07904dc079a3f36284f8e99a/assets/controller.yaml#L60) which will parse it's values and create a cloud.conf for the driver.

assets/csidriveroperators/azure-disk/08_deployment.yaml
```diff
@@ -52,6 +52,10 @@ spec:
           requests:
             memory: 50Mi
             cpu: 10m
+        volumeMounts:
```
RomanBednar, line 55: @jstuever I'm not sure we need to mount the token here for the operator, as it does not authenticate with Azure afaik. Shouldn't we mount the token into the [controller driver](https://github.com/openshift/azure-disk-csi-driver-operator/blob/f71f1449938a1dde07904dc079a3f36284f8e99a/assets/controller.yaml#L74) instead?

PR #364, opened 2023-04-20
CCO-319: Add serviceAccountNames to azure-disk and azure-file credentials

jstuever:
Add the necessary serviceAccountNames to the azure-disk-csi-driver-operator and azure-file-csi-driver-operator credentials requests to enable Azure workload identity tokens.

### DIFF
--- a/manifests/03_credentials_request_azure.yaml
+++ b/manifests/03_credentials_request_azure.yaml
@@ -8,6 +8,9 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     capability.openshift.io/name: Storage
 spec:
+  serviceAccountNames:
+  - azure-disk-csi-driver-operator
+  - azure-disk-csi-driver-controller-sa
   providerSpec:
     apiVersion: cloudcredential.openshift.io/v1
     kind: AzureProviderSpec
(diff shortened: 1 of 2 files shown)


RomanBednar:
cc @openshift/storage for review - we need approved label before submitting the TRT form I believe.

manifests/03_credentials_request_azure.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     capability.openshift.io/name: Storage
 spec:
+  serviceAccountNames:
+  - azure-disk-csi-driver-operator
+  - azure-disk-csi-driver-controller-sa
```
RomanBednar, line 13: @jstuever I think we need one more SA here: `azure-disk-csi-driver-node-sa`

Because nodes need to authenticate too in order to mount volumes, on the other hand we probably don't need the operator SA at all.

manifests/03_credentials_request_azure_file.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     capability.openshift.io/name: Storage
 spec:
+  serviceAccountNames:
+  - azure-file-csi-driver-operator
+  - azure-file-csi-driver-controller-sa
```
RomanBednar, line 13: @jstuever Same here but with: `azure-file-csi-driver-node-sa`

jstuever: [Added](https://github.com/openshift/cluster-storage-operator/compare/dc451d06502544be9638817dd50bbc8b9c34540a..8f3e4a31502e311917c81feb9da507c88771f56b)

PR #371, opened 2023-05-23
OCPBUGS-13696: Warn about CBT enabled VMs via vsphere-problem-detector

vr4manta:
Added new alert for CBT Check

### DIFF
--- a/assets/vsphere_problem_detector/12_prometheusrules.yaml
+++ b/assets/vsphere_problem_detector/12_prometheusrules.yaml
@@ -107,7 +107,7 @@ spec:
       - alert: VSphereOpenshiftVmsCBTMismatch
         # Using min_over_time to make sure the metric is `1` for whole 5 minutes.
         # A missed scraping (e.g. due to a pod restart) will result in prometheus re-evaluating the the alerting rule.
-        expr:  min_over_time(vsphere_vm_cbt_checks{cbt=~"enabled"}[5m]) > 0 and ignoring (cbt) min_over_time(vsphere_vm_cbt_checks{cbt=~"disabled"}[5m]) > 0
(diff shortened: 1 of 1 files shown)


gnufied:
cc @jsafrane

assets/vsphere_problem_detector/12_prometheusrules.yaml
```diff
@@ -103,3 +103,20 @@ spec:
           description: |
             The cluster is using vCenter version less than 7.0.2, which is being deprecated by Openshift. A future version of
             Openshift will remove support for vCenter versions lest than 7.0.2 and it is recommended to update your vCenter to the latest version.
+
+      - alert: VSphereOpenshiftVmsCBTMismatch
+        # Using min_over_time to make sure the metric is `1` for whole 5 minutes.
+        # A missed scraping (e.g. due to a pod restart) will result in prometheus re-evaluating the the alerting rule.
+        expr:  min_over_time(vsphere_vm_cbt_checks{cbt=~"enabled"}[5m]) > 0 and ignoring (cbt) min_over_time(vsphere_vm_cbt_checks{cbt=~"disabled"}[5m]) > 0
```
gnufied, line 110: Do you know why we need `ignoring(cbt)` here?

gnufied: IMO `and on()` should work here as well.

vr4manta: Sorry, I just realized we didnt finish this.  Without ignore, it would not work.  I can try on to see if that works.  Is there a reason to avoid ignoring?

vr4manta: I have tested using on() function and it is working.  I've checked in that version based on your recommendation.  Thanks!

PR #373, opened 2023-05-26
WIP: Incorporate vsphere_problem_detector_disabled_alerts into alerts

jsafrane:
Disable alerts when `vsphere_problem_detector_disabled_alerts == 1`.

Using `max()` to remove all labels from all metrics, so only the highest value remains. And then compare a metric (say `vsphere_sync_errors`) `> vsphere_problem_detector_disabled_alerts`.

If `vsphere_problem_detector_disabled_alerts` is set, it will be equal to the metric, so no alert will be generated.

cc @openshift/storage


jsafrane:
WIP: depends on https://github.com/openshift/vsphere-problem-detector/pull/116

jsafrane:
I removed WIP, it can be tested with https://github.com/openshift/vsphere-problem-detector/pull/116. Still, it should be merged *after* the aforementioned PR.

jsafrane:
/close
we decided we will remove the PrometheusRules to disable the alert instead.

PR #368, opened 2023-05-08
STOR-1334: update storage operator to read featuregates from API on standalone OCP

deads2k:
The previous method of using the featuregate object and a potentially out of date hardcoded list is left for hypershift until hypershift is ready to switch.

This allows coordination of featuregates via openshift/api and vendoring to openshift/cluster-config-operator. Every operator on standalone kcp consumes from `oc get featuregates` status.

PR #376, opened 2023-05-30
STOR-1334: update storage operator to read featuregates from API

jsafrane:
A second attempt after https://github.com/openshift/cluster-storage-operator/pull/368. Now read the feature gates from API both in standalone cluster and in hypershift - it has correct `featuregate.status`.

@openshift/storage


pkg/operator/standalone_starter.go
```diff
@@ -33,6 +35,31 @@ func startControllerStandAlone(ctx context.Context, controllerConfig *controller
...
+
+	select {
+	case <-featureGateAccessor.InitialFeatureGatesObserved():
+		featureGates, _ := featureGateAccessor.CurrentFeatureGates()
+		klog.Infof("FeatureGates initialized: knownFeatureGates=%v", featureGates.KnownFeatures())
+	case <-time.After(1 * time.Minute):
+		klog.Errorf("timed out waiting for FeatureGate detection")
+		return fmt.Errorf("timed out waiting for FeatureGate detection")
+	}
+	featureGates, err := featureGateAccessor.CurrentFeatureGates()
```
gnufied, line 58: I feel like `CurrentFeatureGates` should block until it has done syncing the loop. Is 1 minute reasonable time btw?

jsafrane: The select above is the blocking sync to get feature gates. 

With `exit()` after 1 minute, kubelet will start exp. backoff restarting the Pod, so I think it's good enough. A cluster without FeatureGate is not a very healthy one.

gnufied: @deads2k what do you think about call to `CurrentFeatures` blocking . I think that this code leaks internal details of  `featuregateAccess`  object.  We can clean this up in a follow up, if we all agree to it.

duanwei33:
I built standalone and hypershift hosted cluster to verify:
In standalone cluster, after enabling the TechPreviewNoUpgrade, CSIDriverSharedResource is installed successfully shortly.
In hypershift hosted cluster, storage CO is healthy, I was trying to enable the TechPreviewNoUpgrade, but seems the CVO doesn't work well and **CSIDriverSharedResource** is not enabled actually from FeatureGate.status.featureGates as below:
```
spec:
  featureSet: TechPreviewNoUpgrade
status:
  featureGates:
  - disabled:
    - name: AdmissionWebhookMatchConditions
    - name: AzureWorkloadIdentity
    - name: BuildCSIVolumes
    - name: CSIDriverSharedResource
    - name: DynamicResourceAllocation
    - name: EventedPLEG
    - name: ExternalCloudProvider
    - name: ExternalCloudProviderAzure
    - name: ExternalCloudProviderGCP
    - name: GCPLabelsTags
    - name: GatewayAPI
    - name: InsightsConfigAPI
    - name: MachineAPIProviderOpenStack
    - name: MatchLabelKeysInPodTopologySpread
    - name: MaxUnavailableStatefulSet
    - name: NodeSwap
    - name: PDBUnhealthyPodEvictionPolicy
    - name: PrivateHostedZoneAWS
    - name: RetroactiveDefaultStorageClass
    - name: SigstoreImageVerification
    enabled:
    - name: AlibabaPlatform
    - name: OpenShiftPodSecurityAdmission
```
Looks like not storage issue, will take further check on this.

duanwei33:
My bad, the featuregate need to be set in management cluster `hostedcluster.spec.configuration`, and `CSIDriverSharedResource` is installed successfully in the hosted cluster after the setting. So PR looks good based on the test above from QE side.

PR #390, opened 2023-08-01
Remove "get" verb from "events" resource of RBAC for provisioner sidecar

mpatlasov:
Let's keep `manifests/09_sidecar-main_provisioner_role.yaml` in sync with upstream (https://github.com/openshift/csi-external-provisioner/blob/master/deploy/kubernetes/rbac.yaml)

See discussion at https://github.com/openshift/aws-ebs-csi-driver-operator/pull/245.

### DIFF
--- a/manifests/09_sidecar-main_provisioner_role.yaml
+++ b/manifests/09_sidecar-main_provisioner_role.yaml
@@ -19,7 +19,7 @@ rules:
     verbs: ["get", "list", "watch"]
   - apiGroups: [""]
     resources: ["events"]
-    verbs: ["get", "list", "watch", "create", "update", "patch"]
+    verbs: ["list", "watch", "create", "update", "patch"]
   - apiGroups: ["storage.k8s.io"]
     resources: ["csinodes"]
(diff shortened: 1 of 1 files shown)


mpatlasov:
@gnufied , I remember you wrote in https://github.com/openshift/aws-ebs-csi-driver-operator/pull/245:

> I think it is unnecessary to remove get permissions when - list, watch etc is included. I would think that, there might be > code that assumes it has get permissions because it has list permissions etc. It would surprise such users.
> 
> It also doesn't buys much in terms of reducing permission vector of the sidecar/operators.

and after that Jan wrote there recently:

> I think we should use upstream RBAC definitions where possible. If we're going to use custom permissions, then we need to review all code an updated driver to see if we need to modify them. If we use the upstream ones, then it's task of the upstream to keep them up to date and we can blindly copy them.

Feel free to comment/reject here if you believe we have to avoid such changes...

Phaow:
/label qe-approved
CI looks good enough for the minor change.

jsafrane:
/test help
just using this PR to check CI jobs

PR #392, opened 2023-08-07
STOR-1432: hypershift: add independent refs for AWS EBS driver controller images

sjenning:
Need to update this first as the authoritative source for https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/storage/assets/10_deployment-hypershift.yaml


sjenning:
cc @jsafrane @gnufied

PR #396, opened 2023-09-10
OCPBUGS-18132: Add patch for allowing configmap updates via clusterrole

gnufied:
Goes with https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/169

Fixes OCPBUGS-18132

### DIFF
--- a/assets/csidriveroperators/vsphere/06_clusterrole.yaml
+++ b/assets/csidriveroperators/vsphere/06_clusterrole.yaml
@@ -37,6 +37,7 @@ rules:
   resourceNames:
   - extension-apiserver-authentication
   - vmware-vsphere-csi-driver-operator-lock
+  - admin-gates
   resources:
   - configmaps
   verbs:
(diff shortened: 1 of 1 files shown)


assets/csidriveroperators/vsphere/06_clusterrole.yaml
```diff
@@ -309,6 +309,8 @@ rules:
   - get
   - list
   - watch
+  - patch
+  - update
```
jsafrane, line 313: This allows the vSphere CSI driver operator to patch any pod and ConfigMap in any namespace. IMO it's too broad.
It should allow only selected ConfigMap in openshift-config-managed namespace, i.e. a Role could be better. And definitely decoupled from Pods.

gnufied: Moved this to use named configmap. I think that works as well.

duanwei33:
See test on https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/169.

PR #400, opened 2023-09-22
update readme

RomanBednar:


README.md
```diff
@@ -4,3 +4,44 @@ Operator that sets OCP cluster-wide storage defaults.
...
+export RESIZER_IMAGE=quay.io/openshift/origin-csi-external-resizer:latest  
+export SNAPSHOTTER_IMAGE=quay.io/openshift/origin-csi-external-snapshotter:latest  
+export NODE_DRIVER_REGISTRAR_IMAGE=quay.io/openshift/origin-csi-node-driver-registrar:latest  
+export LIVENESS_PROBE_IMAGE=quay.io/openshift/origin-csi-livenessprobe:latest  
+export KUBE_RBAC_PROXY_IMAGE=quay.io/openshift/origin-kube-rbac-proxy:latest  
+export CLUSTER_CLOUD_CONTROLLER_MANAGER_OPERATOR_IMAGE=quay.io/openshift/origin-cluster-cloud-controller-manager-operator:latest
+
+
+# Run the operator via CLI
+./azure-file-csi-driver-operator start --kubeconfig $KUBECONFIG --namespace openshift-cluster-csi-drivers
```
mpatlasov, line 47: `azure-file-csi-driver-operator start` is Quick-start of cluster-storage-operator? This doesn't make much sense to me...

RomanBednar: Thanks for the catch, it's copy paste error ofcourse - fixing.

README.md
```diff
@@ -4,3 +4,44 @@ Operator that sets OCP cluster-wide storage defaults.
...
+
+# Set kubeconfig and obtain desired version
+export KUBECONFIG=<path-to-kubeconfig>
+export OPERATOR_IMAGE_VERSION=$(oc get clusterversion/version -o json | jq '.status.desired.version' | tr -d "\"")
+
+# Set the environment variables
+export AZURE_DISK_DRIVER_OPERATOR_IMAGE=quay.io/openshift/origin-azure-disk-csi-driver-operator:latest
+export AZURE_FILE_DRIVER_OPERATOR_IMAGE=quay.io/openshift/origin-azure-file-csi-driver-operator:latest
+export AZURE_DISK_DRIVER_IMAGE=quay.io/openshift/origin-azure-disk-csi-driver:latest
+export AZURE_FILE_DRIVER_IMAGE=quay.io/openshift/origin-azure-file-csi-driver:latest
```
mpatlasov, line 35: This wouldn't be enough on non-Azure installations. E.g., in case of AWS, corresponding variables are `AWS_EBS_DRIVER_IMAGE` and `AWS_EBS_DRIVER_OPERATOR_IMAGE`.

README.md
```diff
@@ -4,3 +4,44 @@ Operator that sets OCP cluster-wide storage defaults.
...
+
+To build and run the operator locally:
+
+```shell
+# Build the operator
+make
+
+# Set kubeconfig and obtain desired version
+export KUBECONFIG=<path-to-kubeconfig>
+export OPERATOR_IMAGE_VERSION=$(oc get clusterversion/version -o json | jq '.status.desired.version' | tr -d "\"")
```
mpatlasov, line 29: I think we need similar line for `OPERAND_IMAGE_VERSION`.

README.md
RomanBednar: That is correct, each provider has different variables. The idea was to have at least something as an example, but this can be written better. Listing all possible variables is probably not useful so I changed the doc to describe where those variables can be found with some examples.

README.md
RomanBednar: I've tested today with AWS EBS and everything worked fine without having this variable set. Can you share the errors you're getting or ideally which scenario do you think requires this variable?

README.md
mpatlasov: Fully agreed. Now the doc looks fine. Thank you!

README.md
mpatlasov: I tested it also and yes, everything worked fine without having this variable set. However, I think it's better to keep settings in-sync with production cluster. I looked at our sources and discovered that the var has name [`operandImageVersionEnvVarName`](https://github.com/openshift/cluster-storage-operator/blob/master/vendor/github.com/openshift/library-go/pkg/operator/status/version.go#L23). And the latter is exposed by `VersionForOperandFromEnv()` [here](https://github.com/openshift/cluster-storage-operator/blob/master/vendor/github.com/openshift/library-go/pkg/operator/status/version.go#L72-L74). This function is called a few times from CSO `operator_starter.go`. I didn't dig it further, but passing the value of this var to `NewStandaloneDriverStarter()` [here](https://github.com/openshift/cluster-storage-operator/blob/master/pkg/operator/operator_starter.go#L189) looks as it might be important.

RomanBednar: I was looking for it too but did not find actual usage, it's probably not being passed down to the operators at this point - maybe none of them.

In production CSO now sets the versions to `"0.0.1-snapshot"` [here](https://github.com/openshift/cluster-storage-operator/blob/68c252fb109e10ce189a36c2911a0d6a9d273689/manifests/10_deployment.yaml#L63-L66). I don't know about other operators but for Azure this works fine because Azure operator tries to load the env variable [itself](https://github.com/openshift/azure-disk-csi-driver-operator/pull/78/files#diff-0d623dfd885adb20f991bda4c2453aebd732ca6dbb4d1d4be6e79805c3b48de6R108) and when it matches with string for [missing version](https://github.com/openshift/library-go/blob/ab179469ba3812d4c7bca996895603ad18b281fd/pkg/operator/configobserver/featuregates/simple_featuregate_reader.go#L181C5-L181C47) (also [set](https://github.com/openshift/azure-disk-csi-driver-operator/pull/78/files#diff-0d623dfd885adb20f991bda4c2453aebd732ca6dbb4d1d4be6e79805c3b48de6R109) by the operator) featuregate accessor tries to fall back and read from clusterversion as I've shown in the README. That means for testing we could just use this instead of loading the version up front:

```
export OPERATOR_IMAGE_VERSION="0.0.1-snapshot"
export OPERAND_IMAGE_VERSION=$OPERATOR_IMAGE_VERSION
```

That's one use case I know of, not sure if there are others. We can keep it documented like this and update again when we find a scenario where it does not work.

PR #412, opened 2023-10-26
OCPBUGS-22443: manila: Add missing CSI_FSGROUP_POLICY env var

stephenfin:
This was mistakenly introduced to the wrong set of manifests in a previous change to the Manila CSI Driver Operator [1]. Correct this mistake.

/cc @fmount

[1] https://github.com/openshift/csi-driver-manila-operator/commit/660bbb775b09eefc7df4ec9587b70d657f7d31ff

Signed-off-by: Stephen Finucane <stephenfin@redhat.com>

### DIFF
--- a/assets/csidriveroperators/manila/07_deployment.yaml
+++ b/assets/csidriveroperators/manila/07_deployment.yaml
@@ -45,6 +45,8 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.name
+        - name: CSI_FSGROUP_POLICY
+          value: None
         image: ${OPERATOR_IMAGE}
         imagePullPolicy: IfNotPresent
         name: manila-csi-driver-operator


jsafrane:
@stephenfin in https://github.com/openshift/csi-driver-manila-operator/pull/142#discussion_r883454821 @gouthampacha  it's for development and testing only. Is it now intended for cluster admins? You will need some API for that.

/hold

gouthampacha:
> openshift/csi-driver-manila-operator#142 (comment)

still useful only for developers/testing @jsafrane @stephenfin

jsafrane:
The operator already uses a good default when it's not set. I think it makes very little sense to set the env. variable to the default value here.

I am still not sure what do you want. You already have a quick & dirty env. variable for any testing you may need. If you want it usable by cluster admins, then please add a real Manila-specific API field to [ClusterCSIDriver](https://github.com/openshift/api/blob/c90b4c82852ead1fdb58866b685fe75b0b62163d/operator/v1/types_csi_cluster_driver.go#L125) object - it already allows for driver specific configurations.

stephenfin:
tbc, I only added this because I wanted to remove the misleading assets from the Manila CSI Driver Operator. This is the sole addition on that side that is not reflected here, but from what I'm hearing we don't need/want this particular change here. If so, should I just abandon this and proceed with the removal of manifests from `manila-csi-driver-operator`?

jsafrane:
Yeah, I think this PR does not really belong here. And removing manifests from manila-csi-driver-operator is a good thing to do.

PR #385, opened 2023-07-07
OCPBUGS-15940: Revert revert of hypershift cleanup

gnufied:
Fixes OCPBUGS-15940

PR #417, opened 2023-11-09
OCPBUGS-23120: [IBM ROKS] cluster-storage-operator does not set upgradeable=True

dobsonj:
https://issues.redhat.com/browse/OCPBUGS-23120

## Problem

The upgradeable condition used to be set by SnapshotCRDController, but that was removed in https://github.com/openshift/cluster-storage-operator/pull/385/commits/fa9af3aad65b9d0e9c618453825e4defeaad59ac

DefaultStorageClassController only sets upgradeable=True if [unsupportedPlatformError](https://github.com/openshift/cluster-storage-operator/blob/dbb1514dbf9923c56a4a198374cc59e45f9bc0cc/pkg/operator/defaultstorageclass/controller.go#L97-L100).
IBM VPC returns [supportedbyCSIError](https://github.com/openshift/cluster-storage-operator/blob/dbb1514dbf9923c56a4a198374cc59e45f9bc0cc/pkg/operator/defaultstorageclass/controller.go#L207) instead.
The current code assumes if supportedbyCSIError, then [csidriveroperator controller](https://github.com/openshift/cluster-storage-operator/blob/dbb1514dbf9923c56a4a198374cc59e45f9bc0cc/pkg/operator/csidriveroperator/crcontroller.go#L220-L223) will eventually set it.

The problem is if [shouldRunController](https://github.com/openshift/cluster-storage-operator/blob/dbb1514dbf9923c56a4a198374cc59e45f9bc0cc/pkg/operator/csidriveroperator/driver_starter.go#L336) returns false for a driver with supportedbyCSIError. Then csidriveroperator does not start and nothing will set the condition. shouldRunController always returns false for IBM ROKS because of this [StatusFilter](https://github.com/openshift/cluster-storage-operator/blob/dbb1514dbf9923c56a4a198374cc59e45f9bc0cc/pkg/operator/csidriveroperator/csioperatorclient/ibm-vpc-block.go#L17-L27).

## Solution

DefaultStorageClassController has no way to tell if shouldRunController will eventually return false. So in this PR, CSIDriverStarter sets the condition if no controller is started, because we know at that point no other controller will set the condition.

/cc @openshift/storage @jeffnowicki

### DIFF
--- a/pkg/operator/csidriveroperator/driver_starter.go
+++ b/pkg/operator/csidriveroperator/driver_starter.go
@@ -23,6 +23,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/staticresourcecontroller"
 	"github.com/openshift/library-go/pkg/operator/status"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	"k8s.io/apimachinery/pkg/api/errors"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -49,16 +50,17 @@ type driverInterface interface {
 }
 
 type driverStarterCommon struct {
-	commonClients   *csoclients.Clients
-	resyncInterval  time.Duration
-	infraLister     openshiftv1.InfrastructureLister
-	featureGates    featuregates.FeatureGate
-	csiDriverLister storagelister.CSIDriverLister
-	restMapper      *restmapper.DeferredDiscoveryRESTMapper
-	versionGetter   status.VersionGetter
-	targetVersion   string
-	eventRecorder   events.Recorder
-	controllers     []csiDriverControllerManager
+	commonClients     *csoclients.Clients
+	resyncInterval    time.Duration
+	infraLister       openshiftv1.InfrastructureLister
+	featureGates      featuregates.FeatureGate
+	csiDriverLister   storagelister.CSIDriverLister
+	restMapper        *restmapper.DeferredDiscoveryRESTMapper
+	versionGetter     status.VersionGetter
+	targetVersion     string
+	eventRecorder     events.Recorder
+	controllers       []csiDriverControllerManager
+	controllerStarted bool // true if at least one controller has started
 }
 
 type standAloneDriverStarter struct {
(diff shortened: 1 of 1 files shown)


pkg/operator/csidriveroperator/driver_starter.go
```diff
@@ -218,6 +230,14 @@ func (dsrc *driverStarterCommon) sync(ctx context.Context, syncCtx factory.SyncC
 				return err
 			}
 			if !shouldRun {
+				// If shouldRun is false, then the
```
jsafrane, line 233: This branch will be executed for all CSI drivers that do not belong to the current platform, i.e. for most of these: https://github.com/openshift/cluster-storage-operator/blob/813fcbe37e3ad47521601763641999fec920cefa/pkg/operator/operator_starter.go#L203-L214

You can check if this loop actually matched any CSI driver operator and set `Upgradeable=true` only when it did not.

Or, you can always unconditionally create one `StorageOperatorrUpgradeable=true`. If any ClusterCSIDriver then reports `*Upgradeable=False`, [ClusterOperatorStatusController](https://github.com/openshift/cluster-storage-operator/blob/813fcbe37e3ad47521601763641999fec920cefa/pkg/operator/operator_starter.go#L105) will do an union where one `*Upgradeable=false` trumps any number of `*Upgradeable=true`.

dobsonj:
I built https://github.com/openshift/cluster-storage-operator/pull/417/commits/e67877c48a440353c7cf55482b7b9b585c1a0be2 and pushed the image to `http://quay.io/jdobson/cluster-storage-operator:bug23120-fix2`

pkg/operator/csidriveroperator/driver_starter.go
dobsonj: Thanks! I moved the `setUpgradeableTrue` call to after the loop, if none of these controllers have (ever) started.

I'm hesitant to set Upgradeable=true unconditionally because then in the most common case, where at least one controller is started, we may introduce a brief window of time where it's set to true but will soon change to false. Maybe this is benign, but I would like to avoid that flip-flop and set it only if we know those controllers will not run.

dobsonj:
Just one failure in e2e-ibmcloud-csi:
```
: [Jira:"NetworkEdge"] monitor test service-type-load-balancer-availability setup expand_less | 53s
-- | --
{  failed during setup error waiting for replicaset: failed to create RC "service-test": Internal error occurred: resource quota evaluation timed out}
```
/retest

PR #418, opened 2023-11-13
OCPVE-789: annotate credentials request manifests

qJkee:
Add capability annotations to cloud credentials requests

### DIFF
--- a/manifests/03_credentials_request_alibaba.yaml
+++ b/manifests/03_credentials_request_alibaba.yaml
@@ -6,7 +6,7 @@ metadata:
   annotations:
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
-    capability.openshift.io/name: Storage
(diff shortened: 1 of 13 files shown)


jsafrane:
@qJkee what should happen when Storage capability is set and CloudCredential is not on say AWS? Today, the cluster won't install, because the CSI driver does not get its Secret. CI is already complaining in this PR with a cryptic error `AzureDiskDriverControllerServiceControllerAvailable: Waiting for Deployment
AzureDiskCSIDriverOperatorCRAvailable: AzureDiskDriverNodeServiceControllerAvailable: Waiting for the DaemonSet to deploy the CSI Node Service`.

I think we need some code to throw a more reasonable error like "Timed out waiting for Secret XYZ to be present," with an alert / docs about what it means.

jsafrane:
I am afraid documentation is not enough. Nobody reads it. It would be much better to say something in the operator conditions, so it bubbles up to `oc get clusterversion`, alerts, telemetry and so on.

PR #410, opened 2023-10-20
WIP: use granular permissions for gcp credentails request

RomanBednar:
Use permissions documented here: https://docs.openshift.com/container-platform/4.13/installing/installing_gcp/installing-gcp-account.html#minimum-required-permissions-ipi-gcp_installing-gcp-account

PR #427, opened 2023-12-04
CCO-251: TRT-1376: un-revert #410 "CCO-251: use granular permissions for gcp credentails request"

RomanBednar:
Undo this revert: https://github.com/openshift/cluster-storage-operator/pull/426
Origin PR: https://github.com/openshift/cluster-storage-operator/pull/410

Note: After further clarification I've removed the usage of roles altogether in this PR, the permissions are equivalent to origin PR.

### DIFF
--- a/manifests/03_credentials_request_gcp.yaml
+++ b/manifests/03_credentials_request_gcp.yaml
@@ -17,12 +17,39 @@ spec:
   providerSpec:
     apiVersion: cloudcredential.openshift.io/v1
     kind: GCPProviderSpec
-    predefinedRoles:
-      # FIXME: find a replacement for instanceAdmin, since the CSI driver
-      # only needs "compute.instances.[get|attachDisk|DetachDisk]"
-      - roles/compute.instanceAdmin
-      - roles/compute.storageAdmin
-      - roles/iam.serviceAccountUser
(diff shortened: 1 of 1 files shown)


manifests/03_credentials_request_gcp.yaml
```diff
@@ -17,12 +17,39 @@ spec:
...
+      - "resourcemanager.projects.get"
+      - "resourcemanager.projects.list"
+      - "serviceusage.quotas.get"
+      - "serviceusage.services.get"
+      - "serviceusage.services.list"
+      - "iam.serviceAccounts.actAs"
+      - "iam.serviceAccounts.get"
+      - "iam.serviceAccounts.list"
+      - "resourcemanager.projects.get"
+      - "resourcemanager.projects.list"
```
jsafrane, line 52: @RomanBednar how did you choose the permissions? CI does not pass with `Permission compute.instantSnapshots.* is not valid."`

(and no, /retest will really not help here).

RomanBednar: @jsafrane One of the criteria for CCO-251 is to remove usage or roles altogether, I just wanted to test if expanding current roles to permissions would work (the list is from GCE doc). Even though this would not change the permissions as it matches the roles we could remove the roles already if that's the requirement, leaving optimization for later.

I saw you opened another PR for un-reverting the change so I'll close this one, thanks.

PR #437, opened 2024-01-05
STOR-1714: Release leadership on SIGTERM

jsafrane:
When RunOperator() returns no error, library-go waits for leader election goroutine to release the leadership before exiting.

An error returned from RunOperator() takes a shortcut and leads to much faster exit(1), not giving the election goroutine time to release its lease. A subsequent operator pod needs to wait for the old lease to expire, which slows down cluster upgrade and any development / debugging that needs the operator restarted.

cc @openshift/storage


jsafrane:
/label docs-approved
/label px-approved
/label qe-approved
This is not a user facing change & was approved by QE on slack.
/retest-required

PR #442, opened 2024-01-16
STOR-1289: Move vSphere prometheus rules to CSO

jsafrane:
Move rules that are used to sanitize OCP metric for telemetry from cluster-monitoring-operator here. These rules are used only for storage metrics and should live with storage code.

cluster-monitoring-operator PR that removes the rules from their code: https://github.com/openshift/cluster-monitoring-operator/pull/2235

### DIFF
--- a/assets/vsphere_problem_detector/12_prometheusrules.yaml
+++ b/assets/vsphere_problem_detector/12_prometheusrules.yaml
@@ -7,6 +7,23 @@ metadata:
     role: alert-rules
 spec:
   groups:
+    # Rules to remove unnecessary labels before sending metrics through telemetry.
+    - name: vsphere-problem-detector-telemetry.rules
+      rules:
+      - expr: sum by(version)(vsphere_vcenter_info)
+        record: cluster:vsphere_vcenter_info:sum
+      - expr: sum by(version)(vsphere_esxi_version_total)
+        record: cluster:vsphere_esxi_version_total:sum
+      - expr: sum by(hw_version)(vsphere_node_hw_version_total)
+        record: cluster:vsphere_node_hw_version_total:sum
+      - expr: max by(source)(vsphere_topology_tags)
+        record: cluster:vsphere_topology_tags:max
+      - expr: max by(scope)(vsphere_infrastructure_failure_domains)
+        record: cluster:vsphere_infrastructure_failure_domains:max
+      - expr: max by(status)(vsphere_csi_migration{status=~"|LegacyDeprecatedInTreeDriver|CSIWithMigrationDriver"})
+        record: cluster:vsphere_csi_migration:max
+
+    # Alerting rules
     - name: vsphere-problem-detector.rules
       rules:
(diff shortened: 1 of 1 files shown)


jsafrane:
/hold
to coordinate merge with https://github.com/openshift/cluster-monitoring-operator/pull/2235

jsafrane:
cc @openshift/storage

jan--f:
:+1: 
/lgtm

assets/vsphere_problem_detector/12_prometheusrules.yaml
```diff
@@ -7,6 +7,23 @@ metadata:
...
+        record: cluster:vsphere_esxi_version_total:sum
+      - expr: sum by(hw_version)(vsphere_node_hw_version_total)
+        record: cluster:vsphere_node_hw_version_total:sum
+      - expr: max by(source)(vsphere_topology_tags)
+        record: cluster:vsphere_topology_tags:max
+      - expr: max by(scope)(vsphere_infrastructure_failure_domains)
+        record: cluster:vsphere_infrastructure_failure_domains:max
+      - expr: max by(status)(vsphere_csi_migration{status=~"|LegacyDeprecatedInTreeDriver|CSIWithMigrationDriver"})
+        record: cluster:vsphere_csi_migration:max
+
```
chao007, line 25: I could not get value of `cluster:vsphere_csi_migration:max`
curl -G -k -H "Authorization: Bearer $token" \
https://$(oc get route prometheus-k8s-federate -n \
openshift-monitoring -o jsonpath="{.spec.host}")/federate \
--data-urlencode 'match[]={__name__=~"cluster:vsphere_csi_migration:max"}'

`storage.spec` is as below
spec:
  logLevel: Normal
  managementState: Managed
  operatorLogLevel: Normal

and I checked the value of metric `vsphere_csi_migration`
oc -n openshift-monitoring exec -c prometheus prometheus-k8s-0 -- curl -k -H "Authorization: Bearer $token" 'https://prometheus-k8s.openshift-monitoring.svc:9091/api/v1/query?query=vsphere_csi_migration' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    63    0    63    0     0     60      0 --:--:--  0:00:01 --:--:--    60
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": []
  }
}

chao007: @jsafrane Could you help to double check? Do we need this rule in 4.16?

jsafrane: @chao007, Good catch, the source metric `vsphere_csi_migration` was used in 4.13 as part of [STOR-1265 ](https://issues.redhat.com/browse/STOR-1265) and then removed in 4.14.

I updated both this PR and https://github.com/openshift/cluster-monitoring-operator/pull/2235 and removed the metric from telemetry.

PR #444, opened 2024-01-29
OCPBUGS-27159: Allow vSphere CSI driver to be disabled

jsafrane:
vmware-vsphere-csi-driver-operator disables the driver when it cannot talk to vCenter, e.g. when it does not have valid credentials. This enables processing of condition `xxxDisabled: true` introduced in https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/218.

@openshift/storage


bertinatto (commented):
/lgtm
and
/hold
for https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/218

jsafrane:
/hold cancel
this PR should be merged first

PR #445, opened 2024-01-29
STOR-1611: remove ibm-vpc-node-label-updater

dobsonj:
https://issues.redhat.com/browse/STOR-1611
Since https://github.com/openshift/ibm-vpc-block-csi-driver-operator/pull/89 merged, ibm-vpc-node-label-updater is no longer used or deployed. We can remove it from CSO and eventually from the payload.
/cc @openshift/storage


dobsonj:
There should be no user impact from these changes
/label docs-approved
/label px-approved
/label qe-approved

PR #448, opened 2024-02-02
OCPBUGS-28765: Remove master node selector from azure-csi-driver-operator deployment

csrwng:
The standalone patch already adds the master selector to the deployment. It should not be in the base yaml.


Patryk-Stefanski:
@ropatil010 who can we reach out to get all the necessary labels to get this merged?

ropatil010:
/label qe-approved
cc: @jsafrane @mpatlasov for docs, px approval

PR #452, opened 2024-02-06
AUTH-481: Add PSa labels to openshift-cluster-storage-operator namespace

liouk:


jsafrane:
I would lgtm, why is it a draft?

liouk:
> I would lgtm, why is it a draft?

@jsafrane there is some ongoing work that blocks this, see https://issues.redhat.com/browse/AUTH-482

We first need to complete work for that issue; putting a hold to this but reverting draft status.

/hold

PR #459, opened 2024-02-20
AUTH-482: set required-scc for openshift workloads

liouk:


liouk:
Pinned SCC to one more workload (deployment of `cluster-storage-operator`).

jsafrane:
/lgtm

Remove WIP if it's ready to merge.

jsafrane:
/label px-approved
/label docs-approved
This is not user-facing change.

jsafrane:
For QE: this should result in no user visible change. It only forces the right SCCs to our components, so we avoid future issues like https://issues.redhat.com/browse/OCPBUGS-6992.

PR #460, opened 2024-02-21
OCPBUGS-28748: log default storage class names when collector is triggered

RomanBednar:
It is more helpful to show which storage classes were detected as default. There's an alert when more than one default is present and this message might help debug such cases in the future.

Message example:
```
I0221 16:55:09.385977   65866 starter.go:63] Current default StorageClass count: 2 ([thin-csi thin-csi2])
```

### DIFF
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -53,12 +53,14 @@ func countStorageClasses(clients *csoclients.Clients) {
 				klog.Fatalf("Failed to get existing storage classes: %s", err)
 			}
 			defaultSCCount := 0
+			var defaultSCNames []string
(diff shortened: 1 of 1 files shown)


jsafrane:
> NOTE: prometheus collector does not seem to handle log levels too well and I could not see the message when using .V(4) - removing it.

That's odd, I can see the log line just fine at level 4:
```
I0222 16:39:37.205256       1 starter.go:61] Current default StorageClass count 1
```

jsafrane:
Please put `V(4)` back. We can argue if V(2) would be more  appropriate, but no log level is IMO bad.

RomanBednar:
I see, must have been a mistake while testing - corrected.

PR #477, opened 2024-05-30
MON-2065: manifests: add k8s storage rules

jan--f:
These rules have until now lived in
https://github.com/openshift/cluster-monitoring-operator, where they where generated from
https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/de834e9a291b49396125768f041e2078763f48b5/alerts/storage_alerts.libsonnet. This however rarely changes so I'm committing the manifest here along with some comments.


jan--f:
@jsafrane is anything else needed for this to move forward?

jsafrane:
/label px-approved
/label docs-approved
This PR does not introduce any user-facing change.

We still need QE to test this PR before merge, I poked them on slack.

duanwei33:
Verified with the open PR along with https://github.com/openshift/cluster-monitoring-operator/pull/2377, I was able to trigger the activation and deactivation of KubePersistentVolumeFillingUp/KubePersistentVolumeInodesFillingUp alerts.

PR #482, opened 2024-06-18
STOR-1593: add patch role to PVs for provisioner

rhrmo:
It is required to update provisioner permissions (add patch permission) because of https://github.com/kubernetes-csi/external-provisioner/pull/1155, which we are Rebasing in https://github.com/openshift/csi-external-provisioner/pull/98. This PR needs to be merged before the rebase one.

/cc @openshift/storage


bertinatto:
/label docs-approved
/label qe-approved
/label px-approved

PR #493, opened 2024-07-23
CNV-44992: HCP custom tolerations integration

davidvossel:
The HostedCluster object for hypershift recently got a [new feature](https://github.com/openshift/hypershift/pull/4339) that allows users creating HCP clusters to set custom tolerations on their HCP pods. The CSO needs to integrate this feature in the same way it integrated the NodeSelector feature for HCP.


ropatil010:
Additional test excuted: 
Make all the Worker nodes as Non Schedulable. Delete the pods and check the pods get deployed on Master nodes.

PR #494, opened 2024-07-26
OCPBUGS-37488: Pass TOOLS_IMAGE env var to AWS EBS CSI Driver operator

bertinatto:
CC @openshift/storage

### DIFF
--- a/manifests/image-references
+++ b/manifests/image-references
@@ -142,4 +142,7 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-powervs-block-csi-driver:latest
-
+  - name: tools
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-tools:latest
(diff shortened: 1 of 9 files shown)


manifests/image-references
```diff
@@ -142,4 +142,7 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-powervs-block-csi-driver:latest
-
+  - name: tools
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-tools:latest
```
Phaow, line 148: Hi @bertinatto , btw could we use a multi-arch image instead? The `quay.io/openshift/origin-tools:latest` only for x86.
```console
$  oc image info quay.io/openshift/origin-tools:latest|grep 'architecture'
               architecture=x86_64
```

bertinatto: Hi @Phaow, I'm not sure this will be a problem because the ART automation will replace this specific image with the `tools` image from the payload. For example, see this arm64 payload image:

```
$ TOOLS_IMAGE=$(oc adm release info registry.ci.openshift.org/ocp-arm64/release-arm64:4.17.0-0.nightly-arm64-2024-07-31-121328 --pullspecs | grep -E '\stools' | awk '{print $2}')

$ oc image info $TOOLS_IMAGE | grep architecture
             architecture=aarch64
```

In addition, we use the `tools` image in [KCMO as well](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/0a9ed573e35751455bfb24842fd2de0b05379ed0/manifests/image-references#L17-L20), so if this was a problem, I'd expect things to have be broken there as well.

Phaow: @bertinatto Got it, thanks for the details!

PR #496, opened 2024-08-08
STOR-1422: Bump(openshift/api): to get EFS volume metrics

bertinatto:
CC @openshift/storage


bertinatto:
/label px-approved
/label docs-approved
/label qe-approved
This just bumps a dependency. We'll get the tags in https://github.com/openshift/csi-operator/pull/255.

PR #497, opened 2024-08-20
OCPBUGS-38183: Use restricted-v2 SCC and remove runAsUser in Hypershift

Patryk-Stefanski:


jsafrane:
/lgtm
/approve
/label px-approved
/label docs-approved
/label qe-approved

This is not (yet) user facing change. The updated 10_deployment-hypershift.yaml needs to be synced to openshift/hypershift repo and should be tested there. It's not possible to test it just with this PR.

PR #499, opened 2024-09-04
OCPBUGS-37663: include azure perms for route table

patrickdillon:
Now that the installer has switched from Terraform to using cluster api to provision Azure clusters, the node subnet is created with an (empty) route table attached. In order to provision a volume on the subnet, permissions to join the route table are required.


duanwei33:
Tested with open PR, azure-file nfs provisioning works, and executed all storage cases which looks good.
/label qe-approved

jsafrane:
/retest-required
sorry, azure is flaky :-(

PR #504, opened 2024-09-16
API-1835: bump openshift/api

bertinatto:
To get  https://github.com/openshift/api/pull/2029. From that PR:
> This will allow coordination between different observers watching the various generations. For instance, what if you want to track different deployments?

This will allows controllers to set `.status.Generations` using [SSA](https://kubernetes.io/docs/reference/using-api/server-side-apply/), which started being introduced in library-go controllers [here](https://github.com/openshift/library-go/pull/1787/files).

/assign @deads2k


bertinatto:
/label docs-approved
/label px-approved
/label qe-approved

PR #505, opened 2024-09-17
NO-JIRA: create CRDs from openshift/api

bertinatto:
CRD schemas are now created here: https://github.com/openshift/api/pull/2032.

This allso prevents changes like [this](https://github.com/openshift/cluster-storage-operator/commit/076e178df7aa8bd9753a2b5de9aa14542412a757) when adding new TechPreview fields to ClusterCSIDriver CRDs.

CC @openshift/storage


bertinatto:
/label docs-approved
/label px-approved
/label qe-approved

PR #513, opened 2024-10-01
NO-JIRA: Remove OLMOperatorRemovalController

stephenfin:
This is legacy from OpenShift 4.6 and should no longer be needed. Remove the final use of the OLM metadata from the Manila driver operator config and remove the controller itself.


jsafrane (commented):
/lgtm
/approve
/label px-approved
/label docs-approved
/label qe-approved

PR #522, opened 2024-10-15
OCPBUGS-41365: add KUBE_RBAC_PROXY_CONTROL_PLANE_IMAGE to other neede…

Patryk-Stefanski:
add KUBE_RBAC_PROXY_CONTROL_PLANE_IMAGE to azure-disk and azure-file hyper shift deployments


gnufied:
This would need a follow up PR in `csi-operator` repo also..

Patryk-Stefanski:
@gnufied  I already made this pr -> https://github.com/openshift/csi-operator/pull/289/files , My understanding is that this hook will use the envar set in the deployment_azure-disk-csi-driver-operator and  deployment_azure-file-csi-driver-operator and override the image in all the containers that have a substring of "kube-rbac". Please do correct me if my understanding is wrong.

PR #531, opened 2024-10-29
NO-JIRA: Remove kustomize-related annotations from generated assets

stephenfin:
These provide no value once we have run kustomize and could be confusing. Remove them.

Yes, this is a nit but I think it's helpful.

### DIFF
--- a/assets/csidriveroperators/aws-ebs/hypershift/guest/kustomization.yaml
+++ b/assets/csidriveroperators/aws-ebs/hypershift/guest/kustomization.yaml
@@ -17,3 +17,11 @@ patches:
         name: PLACEHOLDER
     target:
       annotationSelector: "storage.openshift.io/remove-from=guest"
+  # remove these annotations as they're just noise post-kustomization
+  # note that '~1' is the escaped form of '/'
+  # https://datatracker.ietf.org/doc/html/rfc6901
+  - target:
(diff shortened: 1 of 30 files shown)


assets/csidriveroperators/aws-ebs/hypershift/guest/kustomization.yaml
```diff
@@ -17,3 +17,9 @@ patches:
         name: PLACEHOLDER
     target:
       annotationSelector: "storage.openshift.io/remove-from=guest"
+  # remove these annotations as they're just noise post-kustomization
+  - target:
+      annotationSelector: "storage.openshift.io/remove-from=mgmt"
+    patch: |
+      - op: "remove"
+        path: "/metadata/annotations/storage.openshift.io~1remove-from"
```
gnufied, line 25: Can you add a comment here that `~1` is being used because, using `/` will not work for JSON path  and hence `~1` is replaced version of `/`'? And lets do this for all the removals.

stephenfin: Done

stephenfin:
/test e2e-ibmcloud-csi

That job is clearly unrelated since we're not touching the IBM assets, but let's get a green build all the same.

PR #544, opened 2024-11-27
OCPBUGS-45737: Remove unecessary resources from the guest cluster

MaysaMacedo:
In hypershift the guest has no operator running, so no resources
related to it should be created there. This commit removes the creation
of service account, role and role binding.


mandre:
That makes sense to me, and the CI seems to agree. Is this ready for reviews?

PR #551, opened 2025-01-06
OCPBUGS-45073: Support HCP labels

muraee:
HCP is adding a new API field to allow users creating HCP clusters to set custom labels on all HCP pods https://github.com/openshift/hypershift/pull/5114. cluster-storage-operator needs to integrate this feature in the same way it integrated the NodeSelector feature for HCP.


muraee:
cc @jsafrane

PR #565, opened 2025-03-07
OCPBUGS-55182: Add required Microsoft.Network/natGateways/join/action permission

cadenmarchese:
According to linked access checks in Azure, if the user has configured a nat gateway on their subnet, the network resource provider will also check for Microsoft.Network/natGateways/join/action when FSO attempts to execute a Microsoft.Network/virtualNetworks/subnets/write operation. Since the FSO credentialsrequest requires Microsoft.Network/virtualNetworks/subnets/write, and many users also bring a nat gateway, Microsoft.Network/natGateways/join/action is also required.

### DIFF
--- a/manifests/03_credentials_request_azure_file.yaml
+++ b/manifests/03_credentials_request_azure_file.yaml
@@ -18,6 +18,7 @@ spec:
     permissions:
       - 'Microsoft.Network/networkSecurityGroups/join/action'
       - 'Microsoft.Network/routeTables/join/action'
+      - 'Microsoft.Network/natGateways/join/action'
       - 'Microsoft.Network/virtualNetworks/subnets/read'
       - 'Microsoft.Network/virtualNetworks/subnets/write'
       - 'Microsoft.Storage/storageAccounts/delete'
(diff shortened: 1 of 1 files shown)


jsafrane:
@cadenmarchese can you please post a PR upstream ([here](https://github.com/kubernetes-sigs/azurefile-csi-driver/blob/418c016f01afdd4bf3e4bb106c594101805ca73a/docs/driver-parameters.md?plain=1#L7-L24))? We don't want to diverge from upstream permissions, as it would be just us to maintain them.

cadenmarchese:
@jsafrane thanks, I went ahead and opened a PR there: https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/2495

jsafrane:
Now that the upstream PR is merged, I can see there were some [more additions](https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/2495/files). Do we need them in OCP?

cadenmarchese:
@jsafrane that depends - does OCP support configuring those additional resources (serviceEndpointPolicies, networkIntentPolicies, ipamPools) on the subnet?

cadenmarchese:
Closing as this was addressed in https://github.com/openshift/cluster-storage-operator/pull/576 already. Thanks @RomanBednar

PR #569, opened 2025-03-13
STOR-1881: Role and ClusterRole with too wide permissions

dobsonj:
https://issues.redhat.com/browse/STOR-1881

- **rbac: drop obsolete csi.openshift.io apiGroup**
- **rbac: remove wildcard for vsphere cns.vmware.com apiGroup**

/cc @openshift/storage


duanwei33:
CI results look good, which should be sufficient for merging from QE side.
/label qe-approved

PR #572, opened 2025-03-19
STOR-1882: Role and ClusterRole with redundant permissions

dobsonj:
https://issues.redhat.com/browse/STOR-1882

/cc @openshift/storage

### DIFF
--- a/assets/csidriveroperators/openstack-cinder/base/05_clusterrole.yaml
+++ b/assets/csidriveroperators/openstack-cinder/base/05_clusterrole.yaml
@@ -17,8 +17,14 @@ rules:
 # - read/write openshift-cluster-csi-drivers/cloud-provider-config
 # - read-only kube-system/extension-apiserver-authentication
 # - read/write openstack-cinder-csi-driver-operator-lock
+# - read/write openstack-cinder-csi-driver-trusted-ca-bundle
 - apiGroups:
   - ''
+  resourceNames:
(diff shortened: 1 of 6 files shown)


dobsonj:
/hold again
`level=error msg=ManilaCSIDriverOperatorCRDegraded: ResourceSyncControllerDegraded: configmaps is forbidden: User "system:serviceaccount:openshift-cluster-csi-drivers:manila-csi-driver-operator" cannot create resource "configmaps" in API group "" in the namespace "openshift-manila-csi-driver"`

duanwei33:
Waiting for the green CI and getting notified.

dobsonj:
> level=error msg=ManilaCSIDriverOperatorCRDegraded: ResourceSyncControllerDegraded: configmaps is forbidden: User "system:serviceaccount:openshift-cluster-csi-drivers:manila-csi-driver-operator" cannot create resource "configmaps" in API group "" in the namespace "openshift-manila-csi-driver"

This turns out to be a dead-end. An important excerpt from https://kubernetes.io/docs/reference/access-authn-authz/rbac/ –

`You cannot restrict create or deletecollection requests by their resource name. For create, this limitation is because the name of the new object may not be known at authorization time.`

Manila is somewhat unique in that the operator deploys the driver in a different namespace than the operator, so it's not possible to restrict create requests for configmaps to specific resourceNames. There is no sense in pursuing this change any further.

/close

PR #575, opened 2025-04-09
OCPBUGS-54760: Replace deprecated `node-role` toleration/nodeSelector

stephenfin:
The `node-role.kubernetes.io/master` annotation is deprecated in favour of `node-role.kubernetes.io/control-plane` [[1][1]] [[2][2]] and currently produces a warning when modifying resources via `oc`/`kubectl`.

```
❯ oc scale -n openshift-cluster-storage-operator deployment cluster-storage-operator --replicas 0
Warning: spec.template.spec.nodeSelector[node-role.kubernetes.io/master]: use "node-role.kubernetes.io/control-plane" instead
deployment.apps/cluster-storage-operator scaled
```

Migrate all `tolerations` and `nodeSelector` instances over. While both labels currently co-exist, we opt to update the existing label rather than append a new one since the new label has been present since 1.26. This is achieved like so:

```
❯ sed -i 's/node-role.kubernetes.io\/master/node-role.kubernetes.io\/control-plane/' \
    $(ag node-role -l --ignore generated --ignore vendor)
❯ make update
```

[1]: https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/kubeadm/2067-rename-master-label-taint
[2]: https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-master-taint

### DIFF
--- a/assets/csidriveroperators/aws-ebs/hypershift/mgmt/deployment.patch.yaml
+++ b/assets/csidriveroperators/aws-ebs/hypershift/mgmt/deployment.patch.yaml
@@ -56,7 +56,7 @@ spec:
       tolerations:
         - key: CriticalAddonsOnly
           operator: Exists
-        - key: node-role.kubernetes.io/master
+        - key: node-role.kubernetes.io/control-plane
           operator: Exists
           effect: "NoSchedule"
         - key: hypershift.openshift.io/control-plane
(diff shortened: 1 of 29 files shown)


stephenfin:
> Migrate all tolerations and nodeSelector instances over. While both labels currently co-exist, we opt to update the existing label rather than append a new one since the new label has been present since 1.26.

I think that assumption may be incorrect :sweat_smile: 

/hold

stephenfin:
> > Migrate all tolerations and nodeSelector instances over. While both labels currently co-exist, we opt to update the existing label rather than append a new one since the new label has been present since 1.26.
> 
> I think that assumption may be incorrect 😅

MCO is [not currently setting the `node-role.kubernetes.io/control-plane` label on the node](https://github.com/openshift/machine-config-operator/blob/release-4.19/pkg/controller/node/node_controller.go#L329) so it seems the best we can do here is tolerate it for when it does.

PR #576, opened 2025-04-16
OCPBUGS-54491: sync Azure File permissions with upstream docs

RomanBednar:
If users configure a storage class with `networkEndpointType: privateEndpoint` parameter CSI driver will fail to provision a volume due to missing permissions.

I've cross checked the OCPBUGS-54491 bug report with upstream documentation to see what we miss and also noticed a few more that were also missing upstream when testing the private endpoints manually. After discussing this with upstream the list was extended with the missing ones: https://github.com/kubernetes-sigs/azurefile-csi-driver/pull/2513

This CredentialsRequest object is used when installing or upgrading cluster, by extracting it from payload and passing it to `ccoctl` which creates Azure roles with the permissions listed. This might result in granting more privileges than most users need.

Alternatively we could drop this change and only document the process, which would require additional manual step for users intending to use private endpoints.

### DIFF
--- a/manifests/03_credentials_request_azure_file.yaml
+++ b/manifests/03_credentials_request_azure_file.yaml
@@ -16,10 +16,6 @@ spec:
     apiVersion: cloudcredential.openshift.io/v1
     kind: AzureProviderSpec
     permissions:
-      - 'Microsoft.Network/networkSecurityGroups/join/action'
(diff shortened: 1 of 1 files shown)


jsafrane:
> This might result in granting more privileges than most users need.

I personally prefer following upstream than explaining customers what permissions they need and when.
/lgtm

PR #577, opened 2025-04-22
NO-JIRA: correct cinder operator CredentialsRequest name

Phaow:
-  The cinder csi driver operator CredentialsRequest name uses `openshift-cluster-csi-drivers` seems a bit mislead, correct it to `openstack-cinder-csi-driver-operator` which is more clear which operator uses it.


Phaow:
Hi @jsafrane @stephenfin , minimor typo fix, could you please take a look when you get a chance? Thank you!

PR #579, opened 2025-05-12
STOR-2330: Network Policies for CSI drivers and operators

mpatlasov:
This PR includes basic NPs suitable for the most of CSI drivers and operators.


mytreya-rh:
Kindly note that all library-go based operators by [default serve up on port 8443](https://github.com/openshift/secrets-store-csi-driver-operator/blob/5696f89b383632b0f407837ae12663c55e2ee654/vendor/github.com/openshift/library-go/pkg/config/configdefaults/config_default.go#L46). 
So, it maybe a good idea to leave an ingress to that port on the library-go operators.
/cc @dobsonj 
/cc @PillaiManish

PR #586, opened 2025-06-18
NO-JIRA: Remove ibm-cloud-managed-cleanup yaml

lance5890:
1. these yamls has already been deleted in https://github.com/openshift/cluster-storage-operator/pull/253, cvo will skip these yamls in newer version
2. I notice that some PRs still change these yamls, but that should be useless,such as https://github.com/openshift/cluster-storage-operator/pull/458
3. address https://github.com/openshift/cluster-version-operator/issues/1207

so maybe rm the redundant yaml should be a good option


manifests/ibm-cloud-managed-cleanup.yaml
```diff
@@ -1,98 +0,0 @@
-apiVersion: cloudcredential.openshift.io/v1
-kind: CredentialsRequest
-metadata:
-  name: aws-ebs-csi-driver-operator
-  namespace: openshift-cloud-credential-operator
-  annotations:
-    capability.openshift.io/name: CloudCredential
-    include.release.openshift.io/hypershift: "true"
-    include.release.openshift.io/ibm-cloud-managed: "true"
-    release.openshift.io/delete: "true"
```
lance5890, line 10: cvo will skip these yamls with "release.openshift.io/delete" annotation

and this only works when upgrade from older version

PR #598, opened 2025-07-18
STOR-2394: Bump all deps for 4.20

rhrmo:
Bump all deps for 4.20

cc @openshift/storage


RomanBednar:
nit: a description showing how the bump was done would be nice, like: "Updating all dependencies using `go get -u ./...`"

PR #605, opened 2025-08-12
STOR-2523: hypershift: Signal operands version and parent operator in metadata 

muraee:
Signal rollout to a new version is completed for operands running management side

Adds the release.openshift.io/version annotation to CSI driver operand deployments on the management side. This annotation enables HyperShift to track version rollouts and signal when deployment to a new version is completed. Also adds the hypershift.openshift.io/managed-by label to identify operands managed by this operator.


gnufied:
As discussed on Slack, this PR is insufficient to meet the exit criteria of the EPIC. @muraee do you want to close this PR?

PR #608, opened 2025-08-25
STOR-2576: Collect overall SELinux warning controller metrics

jsafrane:
Collect cluster-wide metric that shows nr. of Pods that would conflict, if they land on the same node. To be sent via telemetry.

Keep just the conflicting property as a label. It's either "SELinuxLabel" or "SELinuxChangePolicy", so the final metric cardinality is 2.


jsafrane:
cc @openshift/storage

PR #611, opened 2025-09-02
OCPBUGS-61161: Fix logging e2e also faster cleanup

gnufied:
Bumps openshift-eng-ext to pull logging fixes.

I have also observed that, if you explicitly delete created pod before deleting the namespace, then namespace gets deleted faster. This should help with cleanup of e2e tests in certain environments.

Fixes https://issues.redhat.com/browse/OCPBUGS-61161

### DIFF
--- a/vendor/github.com/openshift-eng/openshift-tests-extension/pkg/cmd/cmdrun/runsuite.go
+++ b/vendor/github.com/openshift-eng/openshift-tests-extension/pkg/cmd/cmdrun/runsuite.go
@@ -1,8 +1,12 @@
 package cmdrun
 
 import (
+	"context"
 	"fmt"
 	"os"
+	"os/signal"
+	"syscall"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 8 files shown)


test/e2e/storage_performant_policy.go
```diff
@@ -232,31 +258,64 @@ var _ = g.Describe("[sig-storage][OCPFeatureGate:StoragePerformantSecurityPolicy
...
+		_, err := kubeClient.CoreV1().Pods(podNamespace).Get(ctx, podName, metav1.GetOptions{})
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, err
+		}
+		return false, nil
+	})
+	g.GinkgoLogr.Info("Pod deleted successfully", "pod", podName, "namespace", podNamespace)
```
mpatlasov, line 314: This must be printed 3 lines below, after `if err != nil { return fmt.Errorf... }`

gnufied: good catch. fixed.

PR #619, opened 2025-09-17
OCPBUGS-60637: remove static value for runAsUser in securityContext

RomanBednar:
See details in commit message. The following PRs have to be merged first:

- https://github.com/openshift/cluster-storage-operator/pull/616
- https://github.com/openshift/csi-operator/pull/431
- https://github.com/openshift/hypershift/pull/6786

hypershift-e2e-aks should be failing until the above PRs are merged. After they're merged the UID allocation will be more dynamic:

- AKS (ARO HCP) does not have SCC and HyperShift will allocate UID that operands have to reflect in  securityContext.runAsUser 
- in self-managed Azure (or any other OpenShift deployment) that has SCC the operands must **not** set securityContext.runAsUser as it is handled by SCC

/cc @openshift/storage @dobsonj @bryan-cox @enxebre

### DIFF
--- a/assets/csidriveroperators/azure-disk/hypershift/mgmt/deployment.patch.yaml
+++ b/assets/csidriveroperators/azure-disk/hypershift/mgmt/deployment.patch.yaml
@@ -64,6 +64,3 @@ spec:
         - name: guest-kubeconfig
           secret:
             secretName: service-network-admin-kubeconfig
-      securityContext:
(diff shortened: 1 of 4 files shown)


dobsonj:
/lgtm
assuming the required hypershift-e2e-aks jobs passes after the other PR's merge :)

duanwei33:
/verified by https://github.com/openshift/csi-operator/pull/431#issuecomment-3336995066

PR #625, opened 2025-10-02
OCPBUGS-62669: Move metrics to vsphere driver

dfajmon:
Move metrics to the vSphere driver from the vSphere problem detector. Before the Prometheus rules were removed with vSphere problem detector.


jsafrane:
/retest-required
Azure should get fixed by https://github.com/openshift/release/pull/69970 (merged ~30 minutes ago)

PR #635, opened 2025-10-16
WIP: Rework custom deployment controller to library-go DeploymentController

jsafrane:
Rework custom `CommonCSIDeploymentController` / `CSIDriverOperatorDeploymentController` / `HyperShiftDeploymentController` to use the default DeploymentController.

The refactoring is taken in small steps

* Refactor functionality of the controllers to manifest / Deployment hooks, one small piece after another. That's majority of the commits in this PR to make review chunks smaller.
* Once every custom code is a hook, remove the controllers and run DeploymentController with the hooks instead. This commit is huge, but it just removes a lot of code.
* The only code that DeploymentController misses is `PostStartHook`. Hack it into `vendor/`, I'll move it into library-go during review. 
* Fix conditions. Despite comments in the code, the old controllers produced `<short name>Progressing` (such as `GCPPDProgressing`) and `<short name>CSIDriverOperatorDeploymentDegraded` (`GCPPDCSIDriverOperatorDeploymentDegraded`). The new DeploymentController progressing condition have the long prefix instead. So add a stale condition remover with the old  `<short name>Progressing` conditions.

### DIFF
--- a/pkg/operator/csidriveroperator/hypershift_deployment_controller.go
+++ b/pkg/operator/csidriveroperator/hypershift_deployment_controller.go
@@ -65,7 +65,7 @@ func NewHyperShiftControllerDeployment(
 	)
 
 	c, err := deploymentcontroller.NewDeploymentControllerBuilder(
-		csiOperatorConfig.ConditionPrefix,
+		csiOperatorConfig.ConditionPrefix+deploymentControllerName,
 		deploymentBytes,
 		eventRecorder,
 		guestClient.OperatorClient,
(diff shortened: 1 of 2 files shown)


gnufied:
Looks like we broke Storage on hypershift:

>     eventually.go:227:  - incorrect condition: wanted ClusterVersionSucceeding=True, got ClusterVersionSucceeding=False: ClusterOperatorNotAvailable(Cluster operator storage is not available)

vendor/github.com/openshift/library-go/pkg/operator/deploymentcontroller/deployment_controller.go
```diff
@@ -192,6 +199,11 @@ func (c *DeploymentController) ToController() (factory.Controller, error) {
 	).ResyncEvery(
 		time.Minute,
 	)
+
+	if len(c.postStartHooks) > 0 {
+		controller = controller.WithPostStartHooks(c.postStartHooks...)
+	}
+
```
gnufied: Custom changes for WIP stuff?

PR #714, opened 2026-06-24
STOR-3011: Implement CSO upgrade check and Prometheus alerts for SELinuxMount readiness

rvagner78:
## Summary by CodeRabbit

* **New Features**
  * Added SELinux mount GA readiness checks that set the operator’s **Upgradeable** status based on the `selinux-conflicts` ConfigMap.
  * Enabled the readiness logic only when the corresponding feature gate is turned on.
* **Bug Fixes**
  * Updated the readiness alerting to directly trigger **SELinuxMountGAReadinessWorkloadsDetected** (including affected pod count) for improved upgrade guidance.
* **Chores**
  * Added RBAC permissions so the readiness controller can read the required ConfigMap.
* **Tests**
  * Added unit tests for conflict parsing and the controller’s sync behavior.


manifests/09_selinux-mount-readiness-configmap-reader-role.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
jsafrane, line 1: CSO already has cluster-admin in the standalone clusters from [here](https://github.com/openshift/cluster-storage-operator/blob/1715686df95688a690e8071c88112aaf3983f4c8/manifests/08_operator_rbac.yaml#L21). 

IMO this role does not does nothing.

manifests/09_selinux-mount-readiness-configmap-reader-role.yaml
```diff
@@ -0,0 +1,48 @@
...
+  labels:
+    kubernetes.io/cluster-service: "true"
+    addonmanager.kubernetes.io/mode: Reconcile
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+    resourceNames:
+      - selinux-conflicts
```
jsafrane, line 22: something tells me that it's not possible to limit `list` by `resourceName` and coderabbit is actually right

manifests/12_prometheusrules.yaml
```diff
@@ -92,6 +92,18 @@ spec:
...
+        for: 10m
+        labels:
+          severity: warning
+        annotations:
+          summary: "Workloads incompatible with SELinuxMount GA detected"
+          description: |
+            Pods with SELinux volume conflicts were detected. These workloads may fail to start
+            after upgrade when SELinuxMount becomes GA. Query metric
+            selinux_warning_controller_selinux_volume_conflict to list affected pods.
+          runbook_url: https://github.com/openshift/enhancements/blob/master/enhancements/storage/selinuxmount-ga-block-upgrade.md
```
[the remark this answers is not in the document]
jsafrane: coderabbit is right here, we should create a PR in openshift/runbooks. I created a new story https://redhat.atlassian.net/browse/STOR-3027

pkg/operator/selinuxmountreadiness/controller.go
```diff
@@ -0,0 +1,151 @@
...
+const (
+	conditionsPrefix = "SELinuxMountGAReadinessController"
+
+
+	selinuxConflictsConfigMapName = "selinux-conflicts"
+	selinuxConflictsDataKey       = "conflictsPresent"
+
+	// SELinuxMountGAReadinessFeatureGate gates the upgrade readiness check until
+	// openshift/api exposes features.FeatureGateSELinuxMountGAReadiness.
+	SELinuxMountGAReadinessFeatureGate = configv1.FeatureGateName("SELinuxMountGAReadiness")
```
jsafrane: Bump openshift/api once https://github.com/openshift/api/pull/2882 is merged and import the feature gate from there.

pkg/operator/selinuxmountreadiness/controller.go
```diff
@@ -0,0 +1,151 @@
...
+		configMapLister: clients.KubeInformers.
+			InformersFor(csoclients.CloudConfigNamespace).
+			Core().V1().ConfigMaps().Lister().ConfigMaps(csoclients.CloudConfigNamespace),
+		featureGates:  featureGates,
+		eventRecorder: eventRecorder,
+	}
+	return factory.New().WithSync(c.sync).WithSyncDegradedOnError(clients.OperatorClient).WithInformers(
+		clients.OperatorClient.Informer(),
+		clients.KubeInformers.InformersFor(csoclients.CloudConfigNamespace).Core().V1().ConfigMaps().Informer(),
+		clients.ConfigInformers.Config().V1().FeatureGates().Informer(),
```
jsafrane: The controller watches FeatureGate CR, but actually never reads it when it changes. It always uses the `c.featureGates` set at the controller creation.

What we usually do is that we read FeatureGates at the operator start and don't start a whole controller if the controller is not enabled.
And when FeatureGates change, we `exit()` and newly started operator fetches a fresh value. See an [example](https://github.com/openshift/cluster-authentication-operator/blob/ed0d09e6a99743a14b1a48cf131e3e9125c86bf7/pkg/operator/starter.go#L742-L757) and [docs](https://github.com/openshift/enhancements/blob/master/dev-guide/feature-zero-to-hero.md#using-your-new-feature-gate)

pkg/operator/selinuxmountreadiness/controller.go
```diff
@@ -0,0 +1,151 @@
...
+func ConflictsPresent(cm *corev1.ConfigMap) (present bool, found bool) {
+	if cm == nil {
+		return false, false
+	}
+	value, ok := cm.Data[selinuxConflictsDataKey]
+	if !ok {
+		return false, true
+	}
+	return metav1.ConditionStatus(value) == metav1.ConditionTrue, true
+}
```
jsafrane: Is this function used anywhere?

pkg/operator/selinuxmountreadiness/controller.go
jsafrane: I can see CSO already has a `featureGateAccessor`. So just remove the informer here and it should be OK. The accessor will `exit()` when the gates change.

manifests/09_selinux-mount-readiness-configmap-reader-role.yaml
rvagner78: yes, it does make much sense right now, however there is a TODO item  "remove cluster-admin from the operator, then a specific UID won't be needed."  in manifests/10_deployment.yaml, so i guess that is the reason why it was added, to indicate that operator needs just read CM selinux-conflicts. But it can be removed  for now.

pkg/operator/selinuxmountreadiness/controller.go
rvagner78: Ok.

pkg/operator/selinuxmountreadiness/controller.go
rvagner78: Fixed, renamed to conflictsPresentInConfigMap and used in conflictPresent() and also in unit tests.

manifests/12_prometheusrules.yaml
```diff
@@ -88,10 +88,18 @@ spec:
       # Pod with set SELinux context successfuly uses a volume (i.e. "mount -o context" would work).
       - expr: sum by(volume_plugin) (volume_manager_selinux_volumes_admitted_total{volume_plugin !~".*-e2e-.*"})
         record: cluster:volume_manager_selinux_volumes_admitted_total
-      # Number of conflicts reported by the SELinux warning controller, i.e. pods that may conflict, if they land on the same node.
-      # The controller does not emit metrics with other properties than "SELinuxLabel" or "SELinuxChangePolicy". Still, let's filter them to be future proof and keep the cardinality low.
-      - expr: sum by(property) (selinux_warning_controller_selinux_volume_conflict{property =~ "SELinuxLabel|SELinuxChangePolicy"})
-        record: cluster:selinux_warning_controller_selinux_volume_conflict:count
```
jsafrane, line 94: Please don't remove this rule, it's used to sum the failures and sent via telemetry.

rvagner78: Ok, the recording rule for telemetry was restored.

manifests/09_selinux-mount-readiness-configmap-reader-role.yaml
jsafrane: Since we're going to remove the controller in 5.1, I think this RBAC will be useless much earlier than we finally implement fine-grained RBAC roles.

pkg/operator/selinuxmountreadiness/controller.go
jsafrane: It's merged now (finally!)

pkg/operator/operator_starter.go
```diff
@@ -185,6 +186,12 @@ func (ssr *StandaloneStarter) StartOperator(ctx context.Context) error {
 		return err
 	}
 
+	if selinuxmountreadiness.FeatureGateEnabled(ssr.featureGates) {
```
jsafrane:
```suggestion
	if ssr.featureGates.Enabled(the feature gate from openshift/api) {
```

And `selinuxmountreadiness.FeatureGateEnabled` and its unit test becomes useless.
(i would expect both cursor  and coderabbit to spot this)

pkg/operator/operator_starter.go
```diff
@@ -289,6 +296,12 @@ func (hsr *HyperShiftStarter) StartOperator(ctx context.Context) error {
 		return err
 	}
 
+	if selinuxmountreadiness.FeatureGateEnabled(hsr.featureGates) {
```
jsafrane: same as above, call `hsrr.featureGates.Enabled`

pkg/operator/selinuxmountreadiness/controller.go
```diff
@@ -0,0 +1,145 @@
...
+	c := &Controller{
+		operatorClient: clients.OperatorClient,
+		configMapLister: corelisters.NewConfigMapLister(configMapInformer.GetIndexer()).
+			ConfigMaps(csoclients.CloudConfigNamespace),
+		eventRecorder: eventRecorder,
+	}
+	return factory.New().WithSync(c.sync).WithSyncDegradedOnError(clients.OperatorClient).WithInformers(
+		clients.OperatorClient.Informer(),
+		configMapInformer,
+	).ToController("SELinuxMountGAReadinessController", eventRecorder), configMapInformer
```
jsafrane: `client.KubeInformers` already watches namespace `openshift-config`, see how it's created here:
https://github.com/openshift/cluster-storage-operator/blob/3a1c01817f1bb47da6ea1390039acf63842b87ab/pkg/csoclients/csoclients.go#L99-L102

Therefore the whole informer creation here is useless. `NewController` does not need to return any informer and this will work:.

```suggestion
	configMapInformer := clients.KubeInformers.InformersFor(csoclients.CloudConfigNamespace).Core().V1().ConfigMaps()
	c := &Controller{
		operatorClient: clients.OperatorClient,
		configMapLister: configMapInformer.Lister().
			ConfigMaps(csoclients.CloudConfigNamespace),
		eventRecorder: eventRecorder,
	}
	return factory.New().WithSync(c.sync).WithSyncDegradedOnError(clients.OperatorClient).WithInformers(
		clients.OperatorClient.Informer(),
		configMapInformer.Informer(),
	).ToController("SELinuxMountGAReadinessController", eventRecorder)
```

pkg/operator/selinuxmountreadiness/controller.go
```diff
@@ -0,0 +1,145 @@
...
+	return coreinformers.NewFilteredConfigMapInformer(
+		kubeClient,
+		csoclients.CloudConfigNamespace,
+		configMapInformerResync,
+		cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},
+		func(listOptions *metav1.ListOptions) {
+			listOptions.FieldSelector = fields.OneTermEqualSelector("metadata.name", selinuxConflictsConfigMapName).String()
+		},
+	)
+}
```
jsafrane: this should not be necessary at all.

pkg/operator/operator_starter.go
```diff
@@ -185,6 +187,12 @@ func (ssr *StandaloneStarter) StartOperator(ctx context.Context) error {
 		return err
 	}
 
+	if ssr.featureGates.Enabled(features.FeatureGateSELinuxMountGAReadiness) {
+		ctrl, configMapInformer := selinuxmountreadiness.NewController(ssr.commonClients, ssr.eventRecorder)
+		ssr.controllers = append(ssr.controllers, ctrl)
+		selinuxmountreadiness.RunConfigMapInformer(configMapInformer, ctx.Done())
```
jsafrane: `selinuxmountreadiness.RunConfigMapInformer` is not necessary, `csoclients.StartInformers()` below already starts `csoclients.KubeInformers`

jsafrane (commented):
The informer creation seems to be not necessary, I hope the comments below make some sense.

pkg/operator/selinuxmountreadiness/controller.go
jsafrane: I admit, github rendering is not great.

jsafrane:
/payload-job periodic-ci-openshift-release-main-ci-5.0-e2e-gcp-ovn-techpreview
/payload-job periodic-ci-openshift-release-main-ci-5.0-e2e-gcp-ovn

I want to see what the operator does *without* https://github.com/openshift/kubernetes/pull/2671

jsafrane:
/verified-by @jsafrane 
Tested together with https://github.com/openshift/cluster-storage-operator/pull/715 and https://github.com/openshift/kubernetes/pull/2671 in CI (see above) and also manually.
Creation of conflicting Pods marks cluster un-upgradeable, their deletion restores upgradeable.

I tested it also on hypershift + AWS.

PR #720, opened 2026-07-24
OCPBUGS-99492: NetworkPolicies should use numeric ports instead of named ports

dobsonj:
https://redhat.atlassian.net/browse/OCPBUGS-99492

/cc @openshift/storage

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated DNS egress network rules to explicitly allow TCP and UDP traffic on port **5353**.
  * Improved DNS resolution compatibility by switching from named/non-numeric DNS port entries to the standard numeric **5353** port.

### DIFF
--- a/manifests/09_network-policy-allow-to-dns.yaml
+++ b/manifests/09_network-policy-allow-to-dns.yaml
@@ -23,8 +23,8 @@ spec:
           dns.operator.openshift.io/daemonset-dns: default
     ports:
     - protocol: TCP
-      port: dns-tcp
+      port: 5353
     - protocol: UDP
-      port: dns
+      port: 5353
   policyTypes:
   - Egress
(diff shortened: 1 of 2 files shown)


dobsonj:
Apparently coredns uses port 5353:
https://github.com/openshift/cluster-dns-operator/blob/4b8ae49940eefc50fa48da5179e735dd6ccd42d9/pkg/manifests/assets/dns/daemonset.yaml#L28-L33
https://github.com/openshift/cluster-dns-operator/blob/4b8ae49940eefc50fa48da5179e735dd6ccd42d9/pkg/manifests/assets/dns/networkpolicy-allow.yaml#L13-L17

dobsonj:
The e2e-*-csi jobs passed after changing the port number from 53 to 5353
/verified by CI

PR #726, opened 2026-08-21
OCPBUGS-91027: Add KAS-defaulted fields to storage-operator ValidatingAdmissionPolicy manifests

natamm4:
The API server auto-injects default fields into the storage-operator's ValidatingAdmissionPolicy manifests, causing the CVO to detect a perpetual diff and hot-loop Update calls every 3–4.5 minutes.

Adds matchPolicy, namespaceSelector, objectSelector, and scope to 13_validating_admission_policy.yaml and 14_validating_admission_policy_binding.yaml.

## Summary by CodeRabbit

* **Updates**
  * Validation policies now apply consistently across all namespace scopes.
  * Equivalent API resources are matched consistently during admission checks.
  * Empty namespace and object selectors are explicitly supported, ensuring resources are evaluated without requiring additional selector configuration.
  * These updates provide more predictable policy enforcement across both namespaced and cluster-scoped resources.

### DIFF
--- a/manifests/13_validating_admission_policy.yaml
+++ b/manifests/13_validating_admission_policy.yaml
@@ -16,6 +16,10 @@ spec:
       apiVersions: ["v1"]
       operations: ["CREATE", "UPDATE"]
       resources: ["namespaces"]
+      scope: "*"
+    matchPolicy: Equivalent
+    namespaceSelector: {}
+    objectSelector: {}
   variables:
   # Check if fsgroup policy label exists
   - name: hasFsGroupPolicyLabel
--- a/manifests/14_validating_admission_policy_binding.yaml
+++ b/manifests/14_validating_admission_policy_binding.yaml
@@ -17,3 +17,7 @@ spec:
       apiVersions: ["v1"]
       operations: ["CREATE", "UPDATE"]
       resources: ["namespaces"]
+      scope: "*"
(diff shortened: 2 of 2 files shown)


mpatlasov:
/lgtm
/approve
/verified by mpatlasov

Verified that before this PR, the ValidatingAdmissionPolicy openshift-storage-policy-validation and the ValidatingAdmissionPolicyBinding openshift-storage-policy-validation-binding do really have those fields set to the same values as in manifests from this PR.

Also verified that before this PR the logs of CVO are cluttered:
```
$ oc logs -n openshift-cluster-version cluster-version-operator-58bd64fdd8-6fjcx |grep -o 'Updating .*due to diff' |grep storage |wc -l
48

$ oc logs -n openshift-cluster-version cluster-version-operator-58bd64fdd8-6fjcx |grep -o 'Updating .*due to diff' |grep storage |tail -2
Updating ValidatingAdmissionPolicy /openshift-storage-policy-validation due to diff
Updating ValidatingAdmissionPolicyBinding /openshift-storage-policy-validation-binding due to diff
```
And they are clean after PR:
```
$ oc logs -n openshift-cluster-version cluster-version-operator-599b9ddb4c-zg6xq |grep -o 'Updating .*due to diff' |grep storage |wc -l
0
```

PR #741, opened 2026-09-07
STOR-1459: Add fine-grained RBAC roles instead of cluster-admin role for cluster-storage-operator

aculnaig:
## Summary by CodeRabbit

- **Security**
  - Replaced broad administrator access with fine-grained permissions for the storage operator in standard environments.
  - Added scoped access across storage, workload, monitoring, RBAC, and platform resources.
  - Added a nonroot-v2 security context binding for the operator workload.
  - Preserved administrator access for hosted environments while updating associated bindings and cleanup manifests.

- **Documentation**
  - Updated RBAC guidance with scoped permissions, minimum-verb requirements, and migration details.
  - Clarified why the deployment’s specific user ID supports reliable scheduling.


manifests/08_operator_rbac-hypershift.yaml
```diff
@@ -2,11 +2,9 @@
...
 kind: ClusterRoleBinding
 metadata:
-  labels:
-    addonmanager.kubernetes.io/mode: Reconcile
-    kubernetes.io/cluster-service: "true"
   name: cluster-storage-operator-role
 roleRef:
+  apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: cluster-admin
```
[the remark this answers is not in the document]
aculnaig: If you consider that I have created a new file named **`manifests/08_0002_operator_rbac-hypershift-delete.yaml`** that will run beforehand **`manifests/08_operator_rbac-hypershift.yaml`** to delete the **`ClusterRoleBinding`** before the new one, I think your suggestion is not necessary. Can you verify ?

manifests/08_0002_cluster_storage_operator_operator_namespace_role.yaml
```diff
@@ -0,0 +1,77 @@
...
+  - apiGroups: [""]
+    resources: ["services"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+
+  # vSphere-specific secrets (split rules for resourceNames compatibility)
+  # Rule 1: List/watch all secrets in namespace (for informer setup)
+  # Informers require list/watch to discover available secrets
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["list", "watch"]
```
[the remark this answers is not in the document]
aculnaig: Can you elaborate more ?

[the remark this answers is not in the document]
aculnaig: I would not bother for this since the operator does **`list`** + **`watch`** secrets in its own namespace now and not cluster-wide anymore.

manifests/08_0001_cluster_storage_operator_storage_role.yaml
```diff
@@ -0,0 +1,56 @@
...
+  # Required for determining platform capabilities, proxy settings, feature gates, and cluster health
+  - apiGroups: ["config.openshift.io"]
+    resources: ["infrastructures", "proxies", "clusterversions", "featuregates", "clusteroperators", "apiservers"]
+    verbs: ["get", "list", "watch"]
+
+  # StorageClass and CSI driver management (cluster-scoped resources)
+  # Required for managing default StorageClasses and platform-specific CSI driver visibility
+  - apiGroups: ["storage.k8s.io"]
+    resources: ["storageclasses", "csidrivers", "volumeattributesclasses"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```
jsafrane: These two are mostly needed by CSI driver operators, not by CSO itself. At least the comment is misleading.

Do all of the CSI driver operators need the same permissions? If so, it would make sense to make a separate ClusterRole and bind CSO + the CSI driver operators to it, otherwise just fix the comment.

manifests/08_0001_cluster_storage_operator_storage_role.yaml
```diff
@@ -0,0 +1,56 @@
...
+    resources: ["clusterrolebindings"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+
+  # Storage operator custom resources (cluster-scoped)
+  # Required for managing the Storage CR and ClusterCSIDriver CRs for each platform
+  - apiGroups: ["operator.openshift.io"]
+    resources: ["storages", "clustercsidrivers"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+  - apiGroups: ["operator.openshift.io"]
+    resources: ["storages/status", "clustercsidrivers/status"]
```
jsafrane: Does CSO itself need to patch `clustercsidrivers/status`? That belongs to CSI driver operator ClusterRole.

manifests/08_0001_cluster_storage_operator_storage_role.yaml
```diff
@@ -0,0 +1,56 @@
...
+  # Required by StatusSyncer controller to report operator health conditions
+  - apiGroups: ["config.openshift.io"]
+    resources: ["clusteroperators/status"]
+    verbs: ["patch", "update"]
+
+  # Monitoring resources for cluster-wide metrics (cluster-scoped)
+  # Required for operator to create and manage cluster-wide ServiceMonitors and PrometheusRules
+  - apiGroups: ["monitoring.coreos.com"]
+    resources: ["servicemonitors", "prometheusrules"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```
jsafrane: ServiceMonitors and PrometheusRules should be namespaced (but double check). And you already have it in a Role below.

manifests/08_0002_cluster_storage_operator_operator_namespace_role.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+  # Rule 2: Get specific vSphere secrets only (restricted by resourceNames for least privilege)
+  # Only vsphere-cloud-credentials and vsphere-problem-detector-serving-cert are needed
+  - apiGroups: [""]
+    resources: ["secrets"]
+    resourceNames:
+      - "vsphere-cloud-credentials"
+      - "vsphere-problem-detector-serving-cert"
+    verbs: ["get"]
```
jsafrane: the operator can `list` all secrets already (and get their content), it does not make much sense to limit `get` only to few of them. Merge it into the `list` rule above.

manifests/08_0002_cluster_storage_operator_operator_namespace_role.yaml
```diff
@@ -0,0 +1,77 @@
...
+  - apiGroups: ["monitoring.coreos.com"]
+    resources: ["servicemonitors", "prometheusrules"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+
+  # Deployment management for VolumeDataSourceValidator
+  # The VolumeDataSourceValidator controller deploys and manages the volume-data-source-validator
+  # deployment in the operator's own namespace for CSI volume population support
+  - apiGroups: ["apps"]
+    resources: ["deployments", "deployments/scale", "deployments/status"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```
jsafrane: Why the operator needs to have any rule for `deployments/status`? IIRC it does not modify it, and it can read it via generic `deployments` rule.

Similarly, the operator may need `update` + `patch`  on `deployments/scale`, but not create/delete/get/list/watch.

manifests/08_0002_cluster_storage_operator_operator_namespace_role.yaml
```diff
@@ -0,0 +1,77 @@
...
+  - apiGroups: [""]
+    resources: ["serviceaccounts"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+
+  # Service management for metrics endpoint
+  - apiGroups: [""]
+    resources: ["services"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+
+  # vSphere-specific secrets (split rules for resourceNames compatibility)
```
jsafrane:
```suggestion
  # vsphere-problem-detector secrets (split rules for resourceNames compatibility)
```

manifests/08_0003_cluster_storage_operator_csi_drivers_namespace_role.yaml
```diff
@@ -0,0 +1,59 @@
...
+    resources: ["events"]
+    verbs: ["create", "update", "patch"]
+
+  # Deployment management for CSI driver operators (DeploymentController)
+  - apiGroups: ["apps"]
+    resources: ["deployments", "deployments/scale"]
+    verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+  - apiGroups: ["apps"]
+    resources: ["deployments/status"]
+    verbs: ["get", "patch", "update"]
```
jsafrane: why does it need to update deployments/status?

manifests/08_0004_cluster_storage_operator_config_namespace_role.yaml
```diff
@@ -0,0 +1,22 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: cluster-storage-operator-config-role
+  namespace: openshift-config
```
jsafrane: You need to check at what runlevel is this namespace created (`oc adm release extract`) and make sure it's before CSO level 50.

manifests/08_0005_cluster_storage_operator_config_managed_namespace_role.yaml
```diff
@@ -0,0 +1,28 @@
...
+  # Read-only access to managed configuration data
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    verbs: ["get", "list", "watch"]
+
+  # RBAC observation for managed cloud provider credentials roles
+  # Read-only access to observe platform-specific RBAC configurations
+  - apiGroups: ["rbac.authorization.k8s.io"]
+    resources: ["roles", "rolebindings"]
+    verbs: ["get", "list", "watch"]
```
jsafrane: This looks suspicious. Why would CSO need this?

manifests/08_0004_cluster_storage_operator_config_namespace_role.yaml
aculnaig: Verified via `oc adm release extract` against the release payload. Both namespaces these Roles target are created by the **cluster-config-operator at runlevel 10**, well before CSO's manifests apply at **runlevel 50**:

| Namespace | Created by | Runlevel |
|---|---|---|
| `openshift-config` | `0000_10_config-operator_01_openshift-config-ns.yaml` | 10 |
| `openshift-config-managed` | `0000_10_config-operator_01_openshift-config-managed-ns.yaml` | 10 |
| CSO manifests (incl. `08_000X` Roles) | `0000_50_cluster-storage-operator_*` | 50 |

CVO applies manifests in ascending runlevel order, so both `openshift-config` and `openshift-config-managed` are guaranteed to exist (10 < 50) before `08_0004` / `08_0005` are applied. Ordering invariant holds — no change needed. Checked both namespaces since `08_0005` has the same cross-namespace dependency.

manifests/08_0001_cluster_storage_operator_storage_role.yaml
aculnaig: Addressed in 1e3281c57

manifests/08_0001_cluster_storage_operator_storage_role.yaml
aculnaig: Addressed in 2a3053156

manifests/08_0001_cluster_storage_operator_storage_role.yaml
aculnaig: Addressed in 2d28547dc

manifests/08_0002_cluster_storage_operator_operator_namespace_role.yaml
aculnaig: Addressed in 24f068f83

manifests/08_0002_cluster_storage_operator_operator_namespace_role.yaml
aculnaig: Addressed in d15d1866c

manifests/08_0002_cluster_storage_operator_operator_namespace_role.yaml
aculnaig: Addressed in 24f068f83

manifests/08_0003_cluster_storage_operator_csi_drivers_namespace_role.yaml
aculnaig: Addressed in 849db2c72

manifests/08_0004_cluster_storage_operator_config_namespace_role.yaml
aculnaig: Addressed in 1a9e2d774

manifests/08_0005_cluster_storage_operator_config_managed_namespace_role.yaml
aculnaig: Addressed in d07f4c7f3

assets/csidriveroperators/aws-ebs/base/06_common_clusterrolebinding.yaml
jsafrane, line 1: This file belongs to assets/csidriveroperators/aws-ebs/base, it's used only by AWS EBS, there is nothing "common" in it.

Similarly, all other ClusterRoleBindings belong to their `base/`.

jsafrane:
/test e2e-aws-csi
???

assets/csidriveroperators/common/clusterrole.yaml
```diff
@@ -0,0 +1,138 @@
...
+  - delete
+  - get
+  - list
+  - update
+  - watch
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - csidrivers
+  - csinodes
```
jsafrane: CSINodes looks suspicious here. Is there a CSI driver operator that needs to access it? I think only the external-attacher and external-provisioner need it and they gets the permission via binding to `openshift-csi-main-*-role`

assets/csidriveroperators/common/clusterrole.yaml
```diff
@@ -0,0 +1,138 @@
...
+  - create
+  - delete
+  - get
+  - list
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
```
jsafrane: I think most of our CSI driver operators don't need to do anything with namespaces. Double-check Manila operator - at some point it was running in a different namespace and I am not sure it's still supported.

assets/csidriveroperators/common/clusterrole.yaml
```diff
@@ -0,0 +1,138 @@
...
+  - nodes
```
jsafrane, line 119: CSI driver operator needs to read nodes (get + list + watch), they should not need editing them (create, patch, update, delete). Please double check.

assets/csidriveroperators/common/clusterrole.yaml
```diff
@@ -0,0 +1,138 @@
...
+  - persistentvolumes
```
jsafrane: CSI driver operator should not need access to PVs. The CSI sidecars get it via `openshift-csi-main-*-role` binding. Maybe there is an operator that needs read the PVs, double check vmware-vsphere-csi-driver-operator?

assets/csidriveroperators/common/clusterrole.yaml
```diff
@@ -0,0 +1,138 @@
...
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - secrets
```
jsafrane: I doubt a CSI driver operator needs to access all Secrets in all namespaces.

assets/csidriveroperators/common/clusterrole.yaml
aculnaig: Good catch on the mechanism — the sidecars *do* get CSINodes via `openshift-csi-main-provisioner-role` and `openshift-csi-main-attacher-role` (both list `csinodes`; see `manifests/09_sidecar-main_{provisioner,attacher}_role.yaml`). But one operator needs it cluster-wide on its own account: **vmware-vsphere**. It both (a) reads CSINodes in operator code (`vspherecontroller.go`, `checks/api_interface.go`) and (b) must *hold* `csinodes` to create its `vmware-vsphere-csi-driver-role` (`assets/rbac/csi_driver_controller_role.yaml`), which the operator applies at runtime — `bind` can't help there, since that's a role creation, not a binding. None of the other 8 operators reference `csinodes` at all.

So I removed it from the common role and moved `csinodes: [get,list,watch]` into the vsphere delta.

assets/csidriveroperators/common/clusterrole.yaml
aculnaig: Confirmed — removed entirely. I checked all 9 operators, Manila included: none call the Namespaces API (get/list/watch/create/…) and none apply a `Namespace` object (Manila only applies a `Secret` in its own operand namespace). The legacy operators keep `namespaces: [get]` in their *namespaced* Role, which is untouched.

assets/csidriveroperators/common/clusterrole.yaml
aculnaig: Agreed, with one exception. Trimmed the common role to `nodes: [get,list,watch]` — no operator needs `create`/`patch`/`delete` (confirmed across all repos). The single write genuinely required is `update`, and only by **ibm-vpc-block**: its `vpc-block-initcontainer-role` (`assets/rbac/initcontainer_role.yaml`) grants `nodes: get/list/update`, so the operator must hold `update` to create it. I moved `nodes: [update]` into the ibm-vpc delta; `get/list/watch` continue to come from common.

assets/csidriveroperators/common/clusterrole.yaml
aculnaig: You're right that the sidecars get PVs via `openshift-csi-main-*-role`, and right to point at vmware-vsphere. Two operators do need PVs cluster-wide:

- **vmware-vsphere** — hold-on-apply: `vmware-vsphere-csi-driver-role` grants PVs full CRUD, so the operator must hold them to create it (notably the vsphere operator code never touches PVs — it's purely the role creation).
- **aws-ebs** — in operator code: the volume-tag controller lists and *updates* PVs cluster-wide (`aws_ebs.go` informer, `aws_ebs_tags_controller.go` `Update`).

So rather than keep PVs in the shared role, I removed them from common and pushed them to the two deltas that actually need them: `persistentvolumes: [get,list,watch,update]` on aws-ebs, full CRUD on vsphere. The other 7 operators no longer get any PV access.

assets/csidriveroperators/common/clusterrole.yaml
aculnaig: Agreed on the principle: no operator reads secrets cluster-wide — every secret read in code is namespaced (cloud credentials in the operand namespace), already covered by the per-driver namespaced Role (e.g. `aws-ebs/base/03_role.yaml` grants secrets full CRUD, namespaced). One caveat that made this rule load-bearing despite looking removable: the azure-file delta already carried `secrets: [create,delete,patch,update]` but relied on the common role for `get/list/watch`, because the azure-file operator must hold secrets **full CRUD** cluster-wide to create `azure-file-csi-driver-role`. So I removed cluster-wide secrets from common and completed the azure-file delta to full CRUD, keeping it self-contained. (smb/samba also has `secrets: get` cluster-wide, but it's a separate *optional* OLM operator with its own SA, not bound to the common role.)

Out of scope for this PR, but worth flagging: `azure-file-csi-driver-role` requesting secrets full CRUD across *all* namespaces looks over-broad in itself — probably a candidate for namespace/`resourceNames` scoping in csi-operator.

aculnaig:
Thanks for the review — addressed all five. Net effect: the common role goes 12 → 11 rules; `csinodes`, `namespaces`, `persistentvolumes` and cluster-wide `secrets` are gone, and `nodes` is now read-only. Everything still required cluster-wide moved into the specific driver's delta instead of being shared across all 9 operators, which tightens least-privilege further.

Two cases aren't outright removable, only relocatable, because the operator must *hold* the rule to create a ClusterRole it applies at runtime (hold-on-apply; `bind` doesn't cover role creation): vsphere (`csinodes`, PVs full CRUD) and azure-file (`secrets` full CRUD). Those went to their deltas.

Validated: `cso-coverage` reports 0 gaps on standalone and hypershift-guest, and a cross-repo hold-on-apply check confirms every driver-operator still holds all rules of the ClusterRoles it creates at runtime (97 permission atoms, 0 uncovered).

assets/csidriveroperators/aws-ebs/base/05_clusterrole.yaml
```diff
@@ -6,327 +6,97 @@ metadata:
...
-  - operator.openshift.io
+  - config.openshift.io
   resources:
-  - clustercsidrivers/status
+  - clusterversions
+  - featuregates
   verbs:
   - get
   - list
   - watch
```
jsafrane, line 16: I see this repeated in aws-ebs, azure-disk, azure-file, gcp-pd.
While  `clusterversion` and `featguregates` may not be used by the others (I did not check them all, at least openstack-cinder does not need it), I think it makes sense to add it to the common role. Reading feature gates and cluster version is not very dangerous that we need to be too much careful about.

assets/csidriveroperators/aws-ebs/base/05_clusterrole.yaml
```diff
@@ -6,327 +6,97 @@ metadata:
...
-  - extension-apiserver-authentication
-  - aws-ebs-csi-driver-operator-lock
+  - discovery.k8s.io
   resources:
-  - configmaps
+  - endpointslices
   verbs:
   - get
   - list
   - watch
```
jsafrane, line 24: Again, this is repeated across several operators. And I am not sure that any operator actually needs EndpointSlices. Can you please check and remove it? Or if it's that common, add it to the common role.

assets/csidriveroperators/aws-ebs/base/05_clusterrole.yaml
```diff
@@ -6,327 +6,97 @@ metadata:
...
-  - delete
-  - patch
-  - update
 - apiGroups:
-  - ''
+  - storage.k8s.io
   resources:
-  - serviceaccounts
+  - volumeattachments/status
   verbs:
```
jsafrane, line 29: I think the operator does not need anything with `volumeattachments/status`. The external-attacher sidecar does and it has its own role for that.

assets/csidriveroperators/aws-ebs/base/05_clusterrole.yaml
```diff
@@ -6,327 +6,97 @@ metadata:
...
-  - update
-  - patch
   - delete
-- apiGroups:
-  - ''
-  resources:
-  - secrets
-  verbs:
   - get
   - list
```
jsafrane, line 49: This looks outdated to me. Once upon a time we had `ConfigMap` based leader election. Now we use `Lease` for that. Can you remove it?

assets/csidriveroperators/aws-ebs/base/05_clusterrole.yaml
```diff
@@ -6,327 +6,97 @@ metadata:
...
-- apiGroups:
-  - storage.k8s.io
-  resources:
-  - storageclasses
-  - csinodes
-  verbs:
-  - create
-  - get
-  - list
   - watch
```
jsafrane, line 66: 1. This belongs to the common ClusterRole
2. No operator should need to update the config map. list/get/watch should be enough.

assets/csidriveroperators/aws-ebs/base/05_clusterrole.yaml
```diff
@@ -6,327 +6,97 @@ metadata:
...
-- apiGroups:
-  - storage.k8s.io
-  resources:
-  - csidrivers
-  verbs:
-  - create
-  - get
-  - list
-  - watch
   - update
```
jsafrane, line 86: 1. It may belong to the common yamls, all driver operators need it.
2. This is too broad. The operator should be able to manipulate namespaced objects only it the openshift-cluster-csi-drivers namespace. I.e. create a Role in that namespace + bind the service account to it.

assets/csidriveroperators/common/clusterrole.yaml
```diff
@@ -0,0 +1,123 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: csi-driver-operator-common-role
+rules:
+- apiGroups:
+  - '*'
+  resources:
+  - events
```
jsafrane, line 9: The operators should be able to manipulate most of namespaced objects only in openshift-cluster-csi-drivers namespace. Create a common Role for it. This applies to lot of resources in this ClusterRole - daemonsets, deployments, replicasets, statefulsets, rolebindings, roles, maybe more.

assets/csidriveroperators/common/clusterrole.yaml
```diff
@@ -0,0 +1,123 @@
...
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - watch
```
jsafrane, line 27: I think this is vSphere specific. Can you check what operators actually create CRDs and perhaps move this just to the vmware-vsphere-csi-driver-operator ClusterRole?

assets/csidriveroperators/aws-ebs/base/05_clusterrole.yaml
jsafrane: This did not render well on github... I mean no operator needs to update ConfigMap named `extension-apiserver-authentication`. Only get/list/watch.

jsafrane: And I just found there is a role `extension-apiserver-authentication-reader` in namespace `kube-system`. Please bind all CSI driver operators (and CSO itself) to that role and remove ClusterRole rules for it.

aculnaig: @jsafrane Good catch — but investigating this one turned up a wrinkle that means it's not a plain removal, and I'd like your call on the shape before I push it.

**The permission is actually needed (not dead like the endpointslices one).** All these operators — and CSO itself — run library-go's default secure-serving (`NewControllerCommandConfig().NewCommand()`, no `DisableServing`, no `SkipInClusterAuthenticationLookup`). That path calls `NewDelegatingAuthenticationOptions().ApplyTo()`, which reads `configmaps/extension-apiserver-authentication` in `kube-system` to set up delegated request-header auth. With `TolerateInClusterLookupFailure=false` (the default), a failed read is **fatal** — so if I just drop the ClusterRole rule without a replacement, the operators fail to start. So binding them to `extension-apiserver-authentication-reader` is exactly right; it's the *replacement*, and it also fixes the current rule being over-broad (full CRUD, cluster-wide, instead of `get/list/watch` on one configmap in `kube-system`).

The complication is *how* to deliver the binding, since CSO applies each driver operator's RBAC from a hardcoded `StaticAssets` list (per driver, per topology). A few options with real trade-offs:

1. **One shared `RoleBinding` in CSO's `manifests/`** (`kube-system`), CVO-applied, `subjects` = CSO's SA + every driver-operator SA. Simplest and escalation-free, one object to review; downside is a central list of all driver SAs and a small hypershift-mgmt variant to sort out.
2. **Per-driver `RoleBinding` assets applied by CSO** (wired into each driver's `StaticAssets`/`MgmtStaticAssets`), plus a CSO binding in `manifests/`, plus giving CSO the escalation right in `kube-system` (hold the reader perms or `bind` on the role) so it can create those bindings. Most per-operator/topology-correct, but a much bigger and riskier change.
3. **Fix CSO only here**, and move the driver operators' switch to the reader binding into csi-operator (the source of truth for their RBAC) as a separate change.

Two related notes:
- The `extension-apiserver-authentication` rule is currently only on aws-ebs, azure-disk, azure-file, gcp-pd, ibm-vpc-block, powervs-block, vsphere and vsphere-problem-detector — **cinder and manila don't have it**, even though they use the same default serving. So there's a pre-existing asymmetry: include them in the binding or leave them as-is?
- csi-operator (upstream) carries neither the rule nor a reader binding today, so whichever way we go, this pattern would be new in CSO — hence option 3 as a convergence-friendly alternative.

My lean is **option 1** (smallest, safest, matches "bind all … to that role"), but it's your/the team's call. Which would you prefer?

assets/csidriveroperators/common/clusterrole.yaml
aculnaig: Agreed. I'll add a namespaced `csi-driver-operator-common-role` Role in `openshift-cluster-csi-drivers` for `events`, apps `daemonsets`/`deployments`/`replicasets`/`statefulsets`, and rbac `roles`/`rolebindings`, bound per-driver (mirroring the existing common ClusterRoleBinding), and drop them from this ClusterRole — keeping `clusterroles`/`clusterrolebindings`, which are inherently cluster-scoped.

Two drivers can't simply move to the namespaced Role, so I'd like your call on one of them before wiring it up:

**vsphere** — the operator applies its own `vmware-vsphere-csi-driver-role` *ClusterRole*, which contains an `events` rule. Creating a ClusterRole requires holding its rules cluster-wide (escalation prevention; `bind` doesn't help create a role), so vsphere must keep `events` cluster-wide. I'll move `events` into the vsphere delta ClusterRole — same approach we used for `csinodes`/`persistentvolumes`/`volumeattachments`. Straightforward.

**manila** — this is the open question. The manila operator manages the driver's Deployment/DaemonSet and RBAC (`prometheus_role`, lease role/binding) in a *separate* namespace, `openshift-manila-csi-driver`, not `openshift-cluster-csi-drivers`. So a common Role in `openshift-cluster-csi-drivers` won't cover it; today it relies on this cluster-wide ClusterRole for `events`/`apps`/`roles`/`rolebindings` there. Two options:

- **(A)** Move manila's `events`/`apps`/`roles`/`rolebindings` into manila's own delta ClusterRole (cluster-wide, manila-only). Unblocks trimming the shared role for the other 8 drivers now and keeps manila working; scoping manila down to `openshift-manila-csi-driver` would be a follow-up.
- **(B)** Add a dedicated manila Role in `openshift-manila-csi-driver` + RoleBinding, wired into manila's static assets (this also requires CSO to hold those permissions in `openshift-manila-csi-driver` to apply the Role).

I lean towards (A) in this PR (smaller, and consistent with how we handled `serviceaccounts` for manila), with (B) as a follow-up — but I'll do (B) directly if you prefer. WDYT?

assets/csidriveroperators/common/clusterrole.yaml
aculnaig: Checked all four operator repos. Good instinct on vSphere — with one nuance: the split is by verb, not wholesale.

**Writes (create/update) — vSphere only.** No operator creates CRDs from assets (only CSO does, for `volumedatasourcevalidator`), and none register a CRD informer. The single operator that needs CRD writes is vSphere: it applies `vmware-vsphere-csi-driver-role`, which lists `customresourcedefinitions: get, create, update`, so its operator must hold those cluster-wide to create that role (escalation prevention). Moved `create`/`update` into the vSphere delta ClusterRole.

**Read (get) — not vSphere specific.** vSphere (`driver_starter.go`), ibm-vpc-block (`starter.go`, gating its `volumesnapshotclass.yaml` conditional controller) and all six csi-operator drivers (`checkForVolumeSnapshotClassCRD`; each ships a `volumesnapshotclass.yaml`) do a `Get` on the `volumesnapshotclasses` CRD before managing VolumeSnapshotClasses — 8 of the 9 operators. So I kept `customresourcedefinitions: get` in the common role rather than duplicate it into eight deltas.

**delete/list/watch/patch — dropped.** No operator deletes a CRD, and none register a CRD informer/lister (the apiextensions informer factory is started but nothing is ever registered on it), so these verbs were dead.

Net: common now grants only `get`; the vSphere delta holds `create`+`update` (with `get` inherited from common) = exactly what its driver-role needs. CSO keeps full `customresourcedefinitions` access in its storage-role, since it applies the volumedatasourcevalidator CRD and both of these ClusterRoles with its own ServiceAccount.

assets/csidriveroperators/aws-ebs/base/06_common_clusterrolebinding.yaml
aculnaig: Good point — the binding is AWS-EBS-specific and belongs in `aws-ebs/base/` (same for every driver: each of the 9 has its binding in its own `base/`).

One nuance on the name, before I rename across the board: `common` here isn't claiming the *binding* is shared — it distinguishes the two ClusterRoleBindings that each driver's `base/` carries:

- `06_clusterrolebinding.yaml` → binds the operator SA to the per-driver **delta** role `aws-ebs-csi-driver-operator-clusterrole`
- `06_common_clusterrolebinding.yaml` → binds the same SA to the **shared** `csi-driver-operator-common-role` (the role factored out across all driver-operators)

So `common` tags the *target role*, not the binding. Dropping it outright would collide with the sibling delta binding above.

Which convention would you prefer? For example: keep the delta binding as `06_clusterrolebinding.yaml` and rename the shared-role one to `06_common_role_clusterrolebinding.yaml` (object name adjusted to match), so it reads as "the binding to the common role". I'll apply whatever you choose across all 9 drivers — filename, object name, generated output, and the `csioperatorclient` asset references.

assets/csidriveroperators/aws-ebs/base/05_clusterrole.yaml
jsafrane: No, I did not ask for a plain removal. Remove the permissions from the operator's ClusterRole / Role, but provide the permissions via (Cluster)RoleBinding of the operator SA to already existing Role `extension-apiserver-authentication-reader`.

jsafrane: I would go with option 2, CSO creates RoleBindings in kube-system for its operands.

assets/csidriveroperators/common/clusterrole.yaml
jsafrane: For manila, yes, A is easier for now. B will come later.

For vSphere, yes, try to localize events only to the driver namespace. I'm checking if the driver really needs all namespaces.

jsafrane: The vSphere CSI driver, as any other CSI driver, needs to access events in all namespaces and we already have a ClusterRole for that: https://github.com/openshift/cluster-storage-operator/blob/main/manifests/09_sidecar-main_provisioner_role.yaml

The vSpher CSI driver operator needs to access events only in its own namespace.

assets/csidriveroperators/aws-ebs/base/05_clusterrole.yaml
aculnaig: Thanks — going with option 2 (CSO creates the RoleBindings in `kube-system` for its operands).

What it looks like:
- For each operand that carried the over-broad rule — aws-ebs, azure-disk, azure-file, gcp-pd, ibm-vpc-block, powervs-block, vsphere, and vsphere-problem-detector — CSO now applies a `RoleBinding` in `kube-system` tying the operator ServiceAccount to the existing `extension-apiserver-authentication-reader` Role, wired into each driver's static asset list (standalone + hypershift-guest).
- Dropped the over-broad `configmaps`/`extension-apiserver-authentication` rule from those operators' ClusterRoles/Roles.
- No escalation change was needed: CSO's storage-role already holds `configmaps` CRUD cluster-wide, which is enough for CSO to create these RoleBindings (no `bind` grant required).

Two things I deliberately left out of this pass — let me know if you'd rather fold them in:
- cinder and manila never carried the rule, so they're not in the binding set (the same pre-existing asymmetry I mentioned above). Easy to add if you want symmetric coverage.
- The hypershift **mgmt** topology (powervs mgmt Role) still has the old rule; I'll handle that as a separate hypershift-mgmt change.

assets/csidriveroperators/common/clusterrole.yaml
aculnaig: Thanks, that settles both.

**manila** — went with (A): its `events`/`apps`/`roles`/`rolebindings` now live in manila's own delta ClusterRole (cluster-wide, manila-only), and the shared role is trimmed for the other 8. Scoping manila down to `openshift-manila-csi-driver` stays a follow-up (your "B later").

**Common role** — added a namespaced `csi-driver-operator-common-role` Role in `openshift-cluster-csi-drivers` (`events`, the apps workloads, and `roles`/`rolebindings`), bound per-driver; the common ClusterRole now keeps only the genuinely cluster-scoped `clusterroles`/`clusterrolebindings`.

**vsphere** — agreed the operator only needs events in its own namespace, and that's what it has now (via the shared namespaced Role plus its own namespaced Role). The one remaining cluster-wide `events` rule on the vsphere operator ClusterRole is purely a hold-on-apply artifact: the operator applies `vmware-vsphere-csi-driver-role`, which today carries its own all-namespaces `events` rule, so the operator has to hold that rule cluster-wide to create it. Your pointer is the clean fix — have the driver get all-namespace events by binding the existing `openshift-csi-main-provisioner-role` (`manifests/09_sidecar-main_provisioner_role.yaml`; the operator already holds `bind` on it) instead of an inline rule in `vmware-vsphere-csi-driver-role`. That's a change in the vSphere driver-operator repo; once it lands I can drop the cluster-wide `events` rule from the vsphere delta here. I'll track it as a follow-up.

assets/csidriveroperators/aws-ebs/extension_apiserver_authentication_rolebinding.yaml
jsafrane, line 1: This file belongs either to `base/` or `standalone/`, depending if it's needed on hypershift or not.

assets/csidriveroperators/common/role.yaml
jsafrane, line 1: This file is too small. I checked these files and they are the same or very similar:
* aws-ebs/base/03_role.yaml
* azure-disk/base/04_role.yaml
* gcp-pd/base/03_role.yaml
* openstack-cinder/base/03_role.yaml
I did not check further. Move the common parts here and leave just the extra perms in CSI driver `base/`.

assets/csidriveroperators/aws-ebs/base/05_clusterrole.yaml
```diff
@@ -5,328 +5,35 @@ metadata:
...
-  - create
-  - get
-  - list
-  - watch
-  - update
-  - delete
 - apiGroups:
   - storage.k8s.io
   resources:
   - volumeattributesclasses
```
jsafrane, line 11: I think the operator itself does not do anything with `volumeattributesclasses`. Only the driver does and there are ClusterRoles for it already (`openshift-csi-provisioner-volumeattributesclass-reader-role` , `openshift-csi-resizer-volumeattributesclass-reader-role`).

(this should apply to all CSI driver operators that have ClusterRole with `volumeattributesclasses`)

assets/csidriveroperators/azure-disk/base/06_clusterrole.yaml
```diff
@@ -6,335 +6,33 @@ metadata:
...
-  resourceNames:
-  - extension-apiserver-authentication
-  - azure-disk-csi-driver-operator-lock
+  - ""
   resources:
   - configmaps
   verbs:
   - get
   - list
   - watch
```
jsafrane, line 23: This is suspicious. Does the operator really ConfigMaps from all namespaces? There is already `azure-disk-csi-driver-operator-secrets-reader-rolebinding` that gives access to `openshift-config` namespace.

assets/csidriveroperators/azure-disk/hypershift/guest/10_rolebinding_config_secrets.yaml
jsafrane, line 1: If it is repeated in `hypershift/guest` and `standalone/`, then it belongs to `base/`. There is a magic annotation ` storage.openshift.io/remove-from: mgmt` that causes kustomize to copy it to `hypershift/guest` and not to `hypershift/mgmt`

(this applies to all *rolebinding_config_secrets.yaml‎)

assets/csidriveroperators/azure-file/base/06_clusterrole.yaml
```diff
@@ -6,331 +6,35 @@ metadata:
...
-  resourceNames:
-  - extension-apiserver-authentication
-  - azure-file-csi-driver-operator-lock
+  - ""
   resources:
   - configmaps
   verbs:
   - get
   - list
   - watch
```
jsafrane, line 15: dtto, this is suspicious

assets/csidriveroperators/gcp-pd/base/05_clusterrole.yaml
```diff
@@ -5,332 +5,46 @@ metadata:
...
-  - get
-  - list
-  - create
-  - watch
-  - delete
 - apiGroups:
   - coordination.k8s.io
   resources:
   - leases
   verbs:
```
jsafrane, line 12: The operator should not access Leases in all namespaces, only in openshift-cluster-csi-drivers.
This applies to all other operators too.

assets/csidriveroperators/ibm-vpc-block/06_clusterrole.yaml
```diff
@@ -3,298 +3,17 @@ kind: ClusterRole
...
-  - get
-  - list
-  - create
-  - watch
-  - delete
 - apiGroups:
   - coordination.k8s.io
   resources:
   - leases
   verbs:
```
jsafrane: dtto, cluster-scoped permissions for leases are likely wrong.

assets/csidriveroperators/openshift-config-managed/secrets-reader-role.yaml
jsafrane, line 1: the file is OK, but the directory name does not fit. Put it into `common/openshift-config-managed-secrets-reader-role.yaml`

assets/csidriveroperators/openshift-config/secrets-reader-role.yaml
jsafrane, line 1: the file is OK, but the directory name does not fit. Put it into common/openshift-config-secrets-reader-role.yaml

assets/csidriveroperators/ibm-vpc-block/06_clusterrole.yaml
jsafrane, line 21: Why the operator needs to access csinodeinfos? It is quite possible that the driver needs it and it's here only to be able to grant it further, but there already is `openshift-csi-main-provisioner-role` with these permissions.

assets/csidriveroperators/openstack-cinder/base/05_clusterrole.yaml
```diff
@@ -6,302 +6,43 @@ metadata:
...
+  - clustercsidrivers
+  - clustercsidrivers/status
   verbs:
-  - get
-  - patch
   - create
-  - list
-  - watch
-  - update
   - delete
```
jsafrane, line 15: The operator should not need to create/delete ClusterCSIDrivers

assets/csidriveroperators/openstack-manila/base/05_clusterrole.yaml
jsafrane, line 1: Cut most of this file into a new Role in `openshift-manila-csi-driver` namespace. The operator does not need to have cluster-wide permissions for most of this.

assets/csidriveroperators/powervs-block/standalone/04_clusterrole.yaml
```diff
@@ -4,301 +4,28 @@ metadata:
...
-  - delete
-- apiGroups:
-  - rbac.authorization.k8s.io
+  - cloudcredential.openshift.io
   resources:
-  - clusterroles
-  - clusterrolebindings
-  - roles
-  - rolebindings
+  - credentialsrequests
```
jsafrane, line 9: The operator should not need any permissions for `credentialsrequests`. CVO creates one unconditionally in `manifests/03_credentials_request_powervs.yaml`

(We do not have CI for powervs!)

assets/csidriveroperators/powervs-block/standalone/04_clusterrole.yaml
```diff
@@ -308,36 +35,15 @@ rules:
   - get
   - list
   - watch
```
jsafrane, line 37: again, the operator should not need `csinodeinfos`. The driver gets them via `openshift-csi-main-provisioner-role`

assets/csidriveroperators/vsphere/06_clusterrole.yaml
```diff
@@ -3,382 +3,145 @@ kind: ClusterRole
...
   resources:
-  - deployments
-  - daemonsets
-  - replicasets
-  - statefulsets
+  - csinodes
   verbs:
   - get
   - list
   - watch
```
jsafrane, line 66: Surprise! This operator actually reads csinodes! So this chunk is correct.

assets/csidriveroperators/vsphere/06_clusterrole.yaml
```diff
@@ -3,382 +3,145 @@ kind: ClusterRole
...
-  - create
-  - update
-  - patch
-  - delete
 - apiGroups:
   - storage.k8s.io
   resources:
   - volumeattachments
```
jsafrane, line 70: I think the operator does not need to access `volumeattachments`, `volumeattachments/status`, `persistentvolumes`, `pods`, `events`, `configmaps` in  all namespaces (but please double check)

manifests/08_0001_cluster_storage_operator_storage_role.yaml
```diff
@@ -0,0 +1,422 @@
...
+    include.release.openshift.io/hypershift: "true"
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    capability.openshift.io/name: Storage
+rules:
+  - apiGroups:
+    - '*'
+    resources:
+    - events
```
jsafrane, line 15: CSO should not need to manage these objects in all namespaces. For the namespaced objects, create two Roles (one for each namespace), for non-namespaced object, double check what is not needed and remove it/

- events
- daemonsets
- deployments
- replicasets
- statefulsets
- leases
- endpointslices?
- volumegroupsnapshotcontents + /status
- prometheusrules
- servicemonitors
- networkpolicies
- poddisruptionbudgets
- roles (create the roles via CVO, i.e. a file in manifests/ in kube-system, openshift-config and other namespaces)
- rolebindings (create role + rolebinding that allows CSO to create RoleBindings in the namespaces above)
- volumesnapshotcontents + status
- volumesnapshots + status 
- volumesnapshots
- volumeattachments + status
- configmaps
- endpoints
- events
- persistentvolumeclaims + status
- pods
- secrets
- serviceaccounts
- services
- credentialsrequests (most probably)

jsafrane: In general, CSO's ClusterRole should be union of all csi-driver-operator ClusterRoles + full access to ClusterCSIDrivers + read + update + patch Storage (+ their statuses). It could make sense to make the union as an extra ClusterRole + have a small ClusterRole for CSO itself.

